PR #11, opened 2018-08-27
Only enable TransactionScopeSuppressFeature if the transport selection if of the current endpoint matches the selected TransportType

danielmarbach:
Suppress behavior could interfere with other transports in multi-endpoint hosting scenarios

### DIFF
--- a/src/Transport/AzureServiceBusTransportInfrastructure.cs
+++ b/src/Transport/AzureServiceBusTransportInfrastructure.cs
@@ -5,6 +5,7 @@
     using System.Text;
     using System.Threading.Tasks;
     using DelayedDelivery;
+    using Features;
     using Microsoft.Azure.ServiceBus;
     using Microsoft.Azure.ServiceBus.Primitives;
     using Performance.TimeToBeReceived;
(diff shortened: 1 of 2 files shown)


src/Transport/Configuration/SettingsKeys.cs
```diff
@@ -13,5 +13,6 @@ static class SettingsKeys
         public const string RuleNameShortener = Base + nameof(RuleNameShortener);
         public const string TransportType = Base + nameof(TransportType);
         public const string CustomTokenProvider = Base + nameof(CustomTokenProvider);
+        public const string Self = Base + "1BA47B92-6987-4B9C-AFF3-B906B9FED0B3";
```
SeanFeldman: `Self` - could you please provide a more meaningful name? Without reading the [comment](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/f33a5b9929a6bd663c8eeaac7359fb6529956607/src/Transport/Receiving/TransactionScopeSuppressFeature.cs) it's hard to determine what this key is for.

danielmarbach: Feel free to change it to your needs. We couldn't come up with a better name. Sorry

bording:
There's an alternative approach to this that we used in the RabbitMQ transport that I think would be a better way to handle this: https://github.com/Particular/NServiceBus.RabbitMQ/pull/472

danielmarbach:
Like that. Simple. Forget that these SettingsExtensions exist

danielmarbach:
@bording @SeanFeldman changed the approach. Have a look

SeanFeldman:
Replaced by #12

PR #12, opened 2018-08-27
Do not auto-activate TransactionScopeSuppressFeature

SeanFeldman:
## Who's affected

Anyone using multi-endpoint host with Azure Service Bus .NET Standard transport and a transport supporing `TransactionScope` [transport transaction](https://docs.particular.net/transports/transactions).

## Symptoms

 Azure Service Bus .NET Standard transport suppresses transaction scope for all transports in multi-endpoint host.

---

Replaces PR #11


SeanFeldman:
Concurrent work FTW :trollface:

PR #24, opened 2018-11-12
Null body

SeanFeldman:
Return empty `byte[]` when a native message has no body.

### DIFF
--- a/src/Tests/Receiving/MessageExtensionsTests.cs
+++ b/src/Tests/Receiving/MessageExtensionsTests.cs
@@ -1,0 +1,30 @@
+﻿namespace NServiceBus.Transport.AzureServiceBus.Tests.Receiving
+{
+    using System;
+    using Configuration;
+    using Microsoft.Azure.ServiceBus;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class MessageExtensionsTests
+    {
+        [Test]
+        public void Should_process_message_without_body()
+        {
+            var message = new Message();
+
+            var body = message.GetBody();
+            Assert.AreEqual(Array.Empty<byte>(), body);
+        }
+
+        [Test]
+        public void Should_process_byte_array_message_without_body()
+        {
+            var message = new Message();
+            message.UserProperties[TransportMessageHeaders.TransportEncoding] = "wcf/byte-array";
+
+            var body = message.GetBody();
+            Assert.AreEqual(Array.Empty<byte>(), body);
+        }
+    }
+}
--- a/src/Transport/Receiving/MessageExtensions.cs
+++ b/src/Transport/Receiving/MessageExtensions.cs
@@ -46,7 +46,7 @@
         {
             if (message.UserProperties.TryGetValue(TransportMessageHeaders.TransportEncoding, out var value) && value.Equals("wcf/byte-array"))
             {
-                return message.GetBody<byte[]>();
+                return message.GetBody<byte[]>() ?? Array.Empty<byte>();
             }
 
             return message.Body ?? Array.Empty<byte>();


src/Tests/Receiving/MessageExtensionsTests.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    using System;
+    using Configuration;
+    using Microsoft.Azure.ServiceBus;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class MessageExtensionsTests
+    {
+        [TestCase]
```
danielmarbach: Test instead of case

src/Tests/Receiving/MessageExtensionsTests.cs
```diff
@@ -0,0 +1,28 @@
...
+        public void Should_not_fail_to_process_octectstream_message_without_body()
+        {
+            var message = new Message();
+            var body = message.GetBody();
+            Assert.AreEqual(Array.Empty<byte>(), body);
+        }
+
+        [TestCase]
```
danielmarbach: Same as above

src/Tests/Receiving/MessageExtensionsTests.cs
```diff
@@ -0,0 +1,28 @@
...
+        public void Should_not_fail_to_process_bytearray_message_without_body()
+        {
+            var message = new Message();
+            message.UserProperties[TransportMessageHeaders.TransportEncoding] = "wcf/byte-array";
+            var body = message.GetBody();
+            Assert.AreEqual(Array.Empty<byte>(), body);
```
danielmarbach: I'd prefer newlines between AAA phases

src/Tests/Receiving/MessageExtensionsTests.cs
```diff
@@ -0,0 +1,28 @@
...
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class MessageExtensionsTests
+    {
+        [TestCase]
+        public void Should_not_fail_to_process_octectstream_message_without_body()
+        {
+            var message = new Message();
+            var body = message.GetBody();
```
danielmarbach: I'd prefer newline between AAA phases

src/Tests/Receiving/MessageExtensionsTests.cs
SeanFeldman: I'm 👍 to separate the message and it's assignments from body and it's test. Test though is coupled to the body which is a system under test.

src/Tests/Receiving/MessageExtensionsTests.cs
SeanFeldman: same as above

SeanFeldman:
Updated based on @danielmarbach feedback

src/Tests/Receiving/MessageExtensionsTests.cs
```diff
@@ -0,0 +1,30 @@
...
+    using System;
+    using Configuration;
+    using Microsoft.Azure.ServiceBus;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class MessageExtensionsTests
+    {
+        [Test]
+        public void Should_not_fail_to_process_octectstream_message_without_body()
```
bording:
```suggestion
        public void Should_process_message_without_body()
```

src/Tests/Receiving/MessageExtensionsTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            var message = new Message();
+
+            var body = message.GetBody();
+            Assert.AreEqual(Array.Empty<byte>(), body);
+        }
+
+        [Test]
+        public void Should_not_fail_to_process_bytearray_message_without_body()
```
bording:
```suggestion
        public void Should_process_byte_array_message_without_body()
```

PR #25, opened 2018-11-20
Rename runtime test queue

SeanFeldman:


src/Transport/Administration/NamespacePermissions.cs
```diff
@@ -23,7 +23,7 @@ public async Task<StartupCheckResult> CanManage()
 
             try
             {
-                await client.QueueExistsAsync("$an-awesome-possum-queue").ConfigureAwait(false);
+                await client.QueueExistsAsync("$nservicebus-verification-queue").ConfigureAwait(false);
```
danielmarbach: Why the dollar in the name?

SeanFeldman:
Queue like that cannot exist :)
The purpose of the method is solely to verify connection string has managed
rights or no.

PR #29, opened 2018-12-06
Release 1.0

Scooletz:
The release of 1.0.0 of this transport.


.gitignore
```diff
@@ -286,3 +286,5 @@ __pycache__/
 *.btm.cs
 *.odx.cs
 *.xsd.cs
+/binaries/*
+/src/Tests/ApprovalFiles/APIApprovals.Approve.received.txt
```
danielmarbach: This could probably be a wildcard but no big deal if we don't change it now

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+  * Clearly describe the issue including steps to reproduce when it is a bug.
+  * If it is a bug make sure you tell us what version you have encountered this bug on.
+* Fork the repository on GitHub
+
+## Running the Acceptance Tests
+ 
+Follow these steps to run the acceptance tests locally:
+
+* Add a new environment variable `Transport.UseSpecific` with the value `AzureServiceBusTransport`
+* Add a new environment variable `AzureServiceBusTransport.ConnectionString` containing a connection string to your Azure storage account
```
danielmarbach: Should be `AzureServiceBus_ConnectionString`
https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/develop/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs#L15

danielmarbach: Also it says Azure storage account which is wrong

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+  * If it is a bug make sure you tell us what version you have encountered this bug on.
+* Fork the repository on GitHub
+
+## Running the Acceptance Tests
+ 
+Follow these steps to run the acceptance tests locally:
+
+* Add a new environment variable `Transport.UseSpecific` with the value `AzureServiceBusTransport`
+* Add a new environment variable `AzureServiceBusTransport.ConnectionString` containing a connection string to your Azure storage account 
+* Add a new environment variable `AzureServiceBusTransport.Topology` and set it to _ForwardingTopology_ to run tests configuring transport with `ForwardingTopology` topology. Don't setup `AzureServiceBusTransport.Topology` environment variable to run tests with `EndpointOrientedTopology` topology
```
danielmarbach: This transport has no concept of a topology. This section was copied over I think

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## Running Unit/Integration Tests
```
danielmarbach: I think this section is obsolete as well. At least we don't use a fallback connection string. The connection string name is wrong as well plus it looks like the tests are not using real connections, see for example https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/develop/src/Tests/Sending/MessageSenderPoolTests.cs#L12

src/CommandLine/Endpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        public static async Task Create(ManagementClient client, CommandArgument name, CommandOption topicName, CommandOption subscriptionName, CommandOption<int> size, CommandOption partitioning)
+        {
+            try
+            {
+                await Queue.Create(client, name, size, partitioning);
+            }
+            catch (MessagingEntityAlreadyExistsException)
+            {
+                Console.WriteLine("Queue already exists, skipping creation");
```
danielmarbach: I think it would have been helpful to include the name of the queue that couldn't be created

src/CommandLine/Endpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+            }
+
+            try
+            {
+                await Topic.Create(client, topicName, size, partitioning);
+            }
+            catch (MessagingEntityAlreadyExistsException)
+            {
+                Console.WriteLine("Topic already exists, skipping creation");
```
danielmarbach: I think it would have been helpful to include the name of the topic that couldn't be created

src/CommandLine/Endpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+            }
+
+            try
+            {
+                await Subscription.Create(client, name, topicName, subscriptionName);
```
danielmarbach: I think it would have been helpful to include the name of the subscription that couldn't be created

src/Transport/Administration/NamespacePermissions.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public NamespacePermissions(ServiceBusConnectionStringBuilder connectionStringBuilder, ITokenProvider tokenProvider)
+        {
+            this.connectionStringBuilder = connectionStringBuilder;
+            this.tokenProvider = tokenProvider;
+        }
+
+        public async Task<StartupCheckResult> CanManage()
+        {
+            var client = new ManagementClient(connectionStringBuilder, tokenProvider);
```
danielmarbach: Does the client need to be closed with `CloseAsync` as the finally block?

https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.management.managementclient.closeasync?view=azure-dotnet

src/Transport/Administration/QueueCreator.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            catch (MessagingEntityAlreadyExistsException)
+            {
+            }
+            // TODO: refactor when https://github.com/Azure/azure-service-bus-dotnet/issues/525 is fixed
+            catch (ServiceBusException sbe) when (sbe.Message.Contains("SubCode=40901.")) // An operation is in progress.
+            {
+            }
+
+            await client.CloseAsync().ConfigureAwait(false);
```
danielmarbach: Should this be in a finally block?

src/Transport/Administration/SubscriptionManager.cs
```diff
@@ -0,0 +1,101 @@
...
+                try
+                {
+                    await client.CreateRuleAsync(topicPath, subscriptionName, rule).ConfigureAwait(false);
+                }
+                catch (MessagingEntityAlreadyExistsException)
+                {
+                }
+            }
+
+            await client.CloseAsync().ConfigureAwait(false);
```
danielmarbach: Should this be in a finally block?

src/Transport/Administration/SubscriptionManager.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            try
+            {
+                await client.DeleteRuleAsync(topicPath, subscriptionName, ruleName).ConfigureAwait(false);
+            }
+            catch (MessagingEntityNotFoundException)
+            {
+            }
+
+            await client.CloseAsync().ConfigureAwait(false);
```
danielmarbach: Should this be in a finally block?

src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+    using System.Collections.Generic;
+    using Configuration;
+    using Microsoft.Azure.ServiceBus;
+    using Microsoft.Azure.ServiceBus.InteropExtensions;
+
+    static class MessageExtensions
+    {
+        public static Dictionary<string, string> GetNServiceBusHeaders(this Message message)
+        {
+            var headers = new Dictionary<string, string>(message.UserProperties.Count);
```
danielmarbach: It would be so nice if we could buffer dictionaries. Probably something for a next version but in theory we know how many concurrent messages we can handle so we could allocate a buffer of dictionaries and then reuse and clear them

src/Transport/Receiving/MessagePump.cs
```diff
@@ -0,0 +1,295 @@
...
+        {
+            try
+            {
+                while (!messageProcessing.IsCancellationRequested)
+                {
+                    await semaphore.WaitAsync(messageProcessing.Token).ConfigureAwait(false);
+
+                    var receiveTask = receiver.ReceiveAsync();
+
+                    ProcessMessage(receiveTask)
```
danielmarbach: Can we simplify this to use non-continuation style?

src/Transport/Receiving/MessagePump.cs
```diff
@@ -0,0 +1,295 @@
...
+                : null;
+        }
+
+        TransportTransaction CreateTransportTransaction(string incomingQueuePartitionKey)
+        {
+            var transportTransaction = new TransportTransaction();
+
+            if (pushSettings.RequiredTransactionMode == TransportTransactionMode.SendsAtomicWithReceive)
+            {
+                transportTransaction.Set((receiver.ServiceBusConnection, receiver.Path));
```
danielmarbach: Is it true that only because of this we need the ValueTuple reference for 4.6.1, see https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/29/files#diff-8a7b07bd3be63680840be24e868393dfR18? If the answer is yes it is a bit overkill I think? We could just as easily build a custom struct and pass that into the transaction

src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public MessageDispatcher(MessageSenderPool messageSenderPool, string topicName)
+        {
+            this.messageSenderPool = messageSenderPool;
+            this.topicName = topicName;
+        }
+
+        public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction transaction, ContextBag context)
+        {
+            // Assumption: we're not implementing batching as it will be done by ASB client
```
danielmarbach: Is the batching support now available or still not yet?

src/Transport/Sending/MessageSenderPool.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        public MessageSender GetMessageSender(string destination, (ServiceBusConnection connection, string path) receiverConnectionAndPath)
+        {
+            var sendersForDestination = senders.GetOrAdd((destination, receiverConnectionAndPath), _ => new ConcurrentQueue<MessageSender>());
+
+            if (!sendersForDestination.TryDequeue(out var sender) || sender.IsClosedOrClosing)
+            {
+                // Send-Via case
+                // TODO: replace with "if (receiverConnectionAndPath != (null, null))" when Resharper inspection 2018.2 is on TeamCity and is set to be default to support C# 7.3 syntax
+                // https://github.com/Particular/Deployment/issues/98
```
danielmarbach: This issue has been closed and therefore the TODO could be adressed

src/CommandLine/Endpoint.cs
SeanFeldman: We probably don't want to bother until this is requested by those that use the tool.
From what we've found, the tool was not very much in [demand](https://www.nuget.org/packages/NServiceBus.Transport.AzureServiceBus.CommandLine/).

src/CommandLine/Endpoint.cs
SeanFeldman: We probably don't want to bother until this is requested by those that use the tool.
From what we've found, the tool was not very much in [demand](https://www.nuget.org/packages/NServiceBus.Transport.AzureServiceBus.CommandLine/).

src/CommandLine/Endpoint.cs
SeanFeldman: We probably don't want to bother until this is requested by those that use the tool.
From what we've found, the tool was not very much in [demand](https://www.nuget.org/packages/NServiceBus.Transport.AzureServiceBus.CommandLine/).

src/Transport/Administration/QueueCreator.cs
SeanFeldman: If not mistaken, we wanted to throw if would encounter one of the exceptions that are not "whitelisted". And if we'd throw, would there be still a point in closing the client in finally?

src/Transport/Administration/SubscriptionManager.cs
SeanFeldman: If not mistaken, we wanted to throw if would encounter one of the exceptions that are not "whitelisted". And if we'd throw, would there be still a point in closing the client in finally?

src/Transport/Administration/SubscriptionManager.cs
SeanFeldman: If not mistaken, we wanted to throw if would encounter one of the exceptions that are not "whitelisted". And if we'd throw, would there be still a point in closing the client in finally?

src/Transport/Receiving/MessageExtensions.cs
SeanFeldman: I might be off, but how much of an overhead we're talking about to justify the added complexity?

src/Transport/Receiving/MessagePump.cs
SeanFeldman: IIRC we needed continuation style. @bording do you recall details?

src/Transport/Receiving/MessagePump.cs
SeanFeldman: We've discussed options, including a custom struct, and decided to go with `ValueTuple` reference for 4.6.1. I suggest raise a separate issue if you feel strongly about it or have an argument that would point out the decision was not the optimal.

src/Transport/Sending/MessageDispatcher.cs
SeanFeldman: Not yet.

src/Transport/Receiving/MessagePump.cs
bording: If we can get the same behavior without it, I don't think it's required. It was the best way we came up with to get the characteristics we wanted here, though.

src/Transport/Receiving/MessageExtensions.cs
bording: I'd definitely want to see some benchmarks before making this sort of change.

src/CommandLine/Endpoint.cs
bording: Seems like it would be worth adding.

src/CommandLine/Endpoint.cs
bording: Seems like it would be worth adding.

src/CommandLine/Endpoint.cs
SeanFeldman: I'm questioning the value, but would suggest to raise a separate issue as a maintenance release not to bog down the release.

src/CommandLine/Endpoint.cs
bording: Seems like it would be worth adding.

src/CommandLine/Endpoint.cs
SeanFeldman: I'm questioning the value, but would suggest to raise a separate issue as a maintenance release not to bog down the release.

src/CommandLine/Endpoint.cs
SeanFeldman: I'm questioning the value, but would suggest to raise a separate issue as a maintenance release not to bog down the release.

SeanFeldman: Replaced with issue #30

src/CommandLine/Endpoint.cs
SeanFeldman: Replaced with issue #30

src/Transport/Administration/NamespacePermissions.cs
SeanFeldman: Fixed in https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/29/commits/df4ee277e63a433f53370ceb7ae6a602141d988f

src/Transport/Sending/MessageSenderPool.cs
SeanFeldman: Fixed in https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/29/commits/cc29335bff3bc7b143e8890120553fc62d018fa6

src/Transport/Receiving/MessageExtensions.cs
SeanFeldman: Feel free to submit a PR when you have good evidence that performance gain justifies the effort. Until then, marking as nice-to-have.

src/Transport/Receiving/MessagePump.cs
SeanFeldman: Raised issue #31

Scooletz:
@danielmarbach Could you take another look? 

I tried to find usages of `AzureServiceBusTransport.ConnectionString` but I failed to do it, so I replaced with `AzureServiceBus_ConnectionString` that looks like the one as it's used in the tests.

SeanFeldman:
@Scooletz  You're correct about `AzureServiceBus.ConnectionString`.It won't be used anywhere in the code. `AzureServiceBus_ConnectionString` is what used.

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+## Running the Acceptance Tests
+
+Follow these steps to run the acceptance tests locally:
+
+* Add a new environment variable `Transport.UseSpecific` with the value `AzureServiceBusTransport`
+* Add a new environment variable `AzureServiceBus_ConnectionString` containing a connection string to your Azure Servie Bus namespace
+
+## Running Unit/Integration Tests
+
+To execute tests under `NServiceBus.AzureServiceBus.Tests`, the following environment variable is required:
```
SeanFeldman:
```suggestion
To execute tests under `NServiceBus.Transport.AzureServiceBus.Tests`, the following environment variable is required:
```

SeanFeldman:
@Scooletz we should hold off with merging until there's some clarity around https://github.com/Azure/azure-service-bus-dotnet/issues/615

SeanFeldman:
Marking as WIP until we resolve the issue with transactions and retries.

PR #32, opened 2018-12-11
Disable ASB retries

### LINKED ISSUE #28: Transaction error
_(Editors note): This is caused by Azure/azure-service-bus-dotnet#615_

Hello 

Now and then I get the following errors. This happens when a surge of events arrive at the queue.

2018-11-29 12:29:30.062 WARN  NServiceBus.Transport.AzureServiceBus.MessagePump Recoverability failed for message with ID 1038f638-88d1-447c-99ab-278ed100eecd. The message will be retried. Exception details: System.InvalidOperationException: Local transactions are not supported with other resource managers/DTC.
   at Microsoft.Azure.ServiceBus.Core.MessageSender.OnSendAsync(IList`1 messageList)
   at Microsoft.Azure.ServiceBus.RetryPolicy.RunOperation(Func`1 operation, TimeSpan operationTimeout)
   at Microsoft.Azure.ServiceBus.RetryPolicy.RunOperation(Func`1 operation, TimeSpan operationTimeout)
   at Microsoft.Azure.ServiceBus.Core.MessageSender.SendAsync(IList`1 messageList)
   at NServiceBus.DelayedRetryExecutor.Retry(IncomingMessage message, TimeSpan delay, TransportTransaction transportTransaction)
   at NServiceBus.RecoverabilityExecutor.DeferMessage(DelayedRetry action, ErrorContext errorContext)
   at NServiceBus.Transport.AzureServiceBus.MessagePump.ProcessMessage(Task`1 receiveTask)

SeanFeldman:
Fixes #28 

Addresses issue https://github.com/Azure/azure-service-bus-dotnet/issues/615 generated by the ASB client.

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -75,11 +75,11 @@
 
             if (tokenProvider == null)
             {
-                receiver = new MessageReceiver(connectionStringBuilder.GetNamespaceConnectionString(), pushSettings.InputQueue, receiveMode, retryPolicy: default, prefetchCount);
+                receiver = new MessageReceiver(connectionStringBuilder.GetNamespaceConnectionString(), pushSettings.InputQueue, receiveMode, retryPolicy: RetryPolicy.NoRetry, prefetchCount);
             }
             else
             {
-                receiver = new MessageReceiver(connectionStringBuilder.Endpoint, pushSettings.InputQueue, tokenProvider, connectionStringBuilder.TransportType, receiveMode, retryPolicy: default, prefetchCount);
+                receiver = new MessageReceiver(connectionStringBuilder.Endpoint, pushSettings.InputQueue, tokenProvider, connectionStringBuilder.TransportType, receiveMode, retryPolicy: RetryPolicy.NoRetry, prefetchCount);
             }
 
             semaphore = new SemaphoreSlim(maxConcurrency, maxConcurrency);
--- a/src/Transport/Sending/MessageSenderPool.cs
+++ b/src/Transport/Sending/MessageSenderPool.cs
@@ -26,17 +26,17 @@
                 // Send-Via case
                 if (receiverConnectionAndPath != (null, null))
                 {
-                    sender = new MessageSender(receiverConnectionAndPath.connection, destination, receiverConnectionAndPath.path);
+                    sender = new MessageSender(receiverConnectionAndPath.connection, destination, receiverConnectionAndPath.path, RetryPolicy.NoRetry);
                 }
                 else
                 {
                     if (tokenProvider == null)
                     {
-                        sender = new MessageSender(connectionStringBuilder.GetNamespaceConnectionString(), destination);
+                        sender = new MessageSender(connectionStringBuilder.GetNamespaceConnectionString(), destination, RetryPolicy.NoRetry);
                     }
                     else
                     {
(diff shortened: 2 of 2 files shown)


bording:
@danielmarbach When @SeanFeldman and I talked about this change, we discussed the idea of keeping the `RetryPolicy` on in some cases, but we decided against it.

For starters, having the retry policy enabled actually masks errors that seem like the would be useful to know about by seeing them in logs. For example, if you're getting an exception because you're throttled, it seems far better to me that you'd actually see that in the logs to know you're being throttled, and can then consider addressing it by moving to a higher capacity namespace. The NSB recoverability will still result in retries, and you'll get the additional info in your logs.

It also seemed strange to be inconsistent in how the transport interacts with ASB based on the transport transaction mode, so we thought it made more sense to keep it off in all cases.

Overall, we didn't see the benefit of adding another retry mechanism when we can rely on NSB to recover from failures, especially since other broker-based transports don't have anything like that.

SeanFeldman:
Additional option is to go strict with this change and then turn `RetryPolicy.Default` on when the client is fixed to differentiate between usages with and w/o transactions. We could track it as an issue dependent on  https://github.com/Azure/azure-service-bus-dotnet/issues/615 and release an update when client is ready. That way we'd be both consistent in our code and allow customers that use `ReceiveOnly` guarantee mode have automatic retries.

danielmarbach:
@bording I can see those arguments and they are certainly valid. I'm trying to wear a different hat hear. It is correct that we'll get additional retries when exceptions are thrown when we execute operations as part of a message being handled. So let's look at this case first. In the throttling case their might be usefulness to see this in the log I agree, but then I'd say the code has to deal with those cases and add corresponding log entries and not use a blunt hammer that also covers throttling. For other short connection losses the pump as well as the dispatcher are no behaving differently. Instead of retrying they immediately throw and eat up more immediate and delayed retries eventually. Since we haven't thoroughly tested this and have been testing the client code with the default policy we are talking here about a behavior change in the transport that might require more testing because at the end we don't know if we are making the right tradeoffs here. There is another angle. Previously we have been flooding our users with too many log details which lead them opening quite a few support cases where we then told them "no its fine, that is all expected". By removing the retry policy we might expose us to more of those scenarios as well.

Then there is the sends outside the message handling pipeline. We are much more likely to throw when the connection drops. That's why I propose to scope it down to SendsWithAtomicReceive only.

yvesgoeleven:
If we disable the built in retry, we should add our own retry policy on top. Transient exceptions are simply very common. We'd exhaust our immediate retry count too quickly

SeanFeldman:
> Transient exceptions are simply very common. We'd exhaust our immediate retry count too quickly

What we've done in the legacy transport was retrying operation ourselves as legacy ASB client didn't always 
retry. To discuss wherever we should or should not have custom retries, I've raised a separate issue #33.

Regarding using, or not, ASB client `RetryPolicy` - we can modify our current code to look at the transport transaction mode. But then we'd need to distinguish between use cases such as an endpoint running in `SendsAtomicWithReceive` yet dispatching in immediate mode, sending with `IMessageSession`, etc. 

Perhaps, rather than trying to make a change such as this one, and then later revert it when ASB client is fixed, we should see if the client can be patched in the next week or two, meanwhile releasing a preview version to help customers that are stuck, and update to the fixed client towards along with `release-1.0` of our transport?

PR #40, opened 2019-01-18
Update test projects to use ASB client to 3.3.0

SeanFeldman:
@Particular/azure-maintainers please review/merge.


SeanFeldman:
My understanding is that we keep the main project locked on the version it used and the test projects are the ones where we force transient dependencies to be on the latest version. Isn't that the case?

If that's not the case, I'll have to refresh my memory and update the PR.

SeanFeldman:
@bording reworked commit comment as per your comment above.

PR #45, opened 2019-01-24
Hotfix-1.0.1

SeanFeldman:


bording:
We should have merged #43 to `master` as a hotfix, but given that didn't happen, this should probably be a `release-1.0.1` branch instead of merging `develop` directly.

SeanFeldman:
Will update.

Issue #67, opened 2019-07-04
.NET Core tool creates subscriptions without rules


danielmarbach:
When using the global tool with 

```
 asb-transport endpoint create Endpoint2
```

it creates the bundle, the queue and the subscription for the endpoint but it creates no rules. It would be nice to be able to specify the event types it wants to subscribe to so that the rules are set up automatically. This would allow running the endpoint without manage rights.

Something like

```
 asb-transport endpoint create Endpoint2 --subscribe-to MyMessage1 --subscribe-to RuleName,MyNamespace.MyMessage
```

PR #68, opened 2019-07-04
Allow rule creation

danielmarbach:
Closes https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/67

Documentation https://github.com/Particular/docs.particular.net/pull/4422

### DIFF
--- a/src/CommandLine/Endpoint.cs
+++ b/src/CommandLine/Endpoint.cs
@@ -57,7 +57,7 @@
             }
             catch (MessagingEntityAlreadyExistsException)
             {
-                Console.WriteLine("Rule does not exists, skipping deletion");
+                Console.WriteLine("Rule does not exist, skipping deletion");
             }
         }
     }
--- a/src/CommandLine/Endpoint.cs
+++ b/src/CommandLine/Endpoint.cs
@@ -36,5 +36,29 @@
                 Console.WriteLine("Subscription already exists, skipping creation");
             }
         }
+
+        public static async Task Subscribe(ManagementClient client, CommandArgument name, CommandOption topicName, CommandOption subscriptionName, CommandOption<int> size, CommandArgument eventType, CommandOption ruleName)
+        {
+            try
+            {
+                await Rule.Create(client, name, topicName, subscriptionName, eventType, ruleName);
+            }
+            catch (MessagingEntityAlreadyExistsException)
+            {
+                Console.WriteLine($"Rule already exists, skipping creation. Verify SQL filter matches '[NServiceBus.EnclosedMessageTypes] LIKE '%{eventType.Value}%'.");
+            }
+        }
+        
+        public static async Task Unsubscribe(ManagementClient client, CommandArgument name, CommandOption topicName, CommandOption subscriptionName, CommandOption<int> size, CommandArgument eventType, CommandOption ruleName)
+        {
+            try
+            {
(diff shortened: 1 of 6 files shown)


src/CommandLine/Endpoint.cs
```diff
@@ -35,6 +35,15 @@ public static async Task Create(ManagementClient client, CommandArgument name, C
...
                 Console.WriteLine("Subscription already exists, skipping creation");
             }
+
+            try
+            {
+                await Rule.Create(client, name, topicName, subscriptionName, subscribeTo);
+            }
+            catch (MessagingEntityAlreadyExistsException)
+            {
+                Console.WriteLine("Rule already exists, skipping creation");
```
SeanFeldman: There might be an issue here, a rule might exist with the same name, but the SQL filter would be different.
As a precaution, I would suggest modifying the message here to indicate a check is required.

SeanFeldman: There might be an issue here, a rule might exist with the same name, but the SQL filter would be different.
As a precaution, I would suggest modifying the message here to indicate a check is required.

SeanFeldman: Also, if there's a request to create multiple rules, and only one of those fails, all the following tasks to create rules might not get executed since creation is done with `Task.WhenAll()`.

src/CommandLine/Program.cs
```diff
@@ -48,10 +48,11 @@ static int Main(string[] args)
                     createCommand.Options.Add(partitioning);
                     var topicName = createCommand.Option("-t|--topic", "Topic name (defaults to 'bundle-1')", CommandOptionType.SingleValue);
                     var subscriptionName = createCommand.Option("-b|--subscription", "Subscription name (defaults to endpoint name) ", CommandOptionType.SingleValue);
+                    var subscribeTo = createCommand.Option("-e|--subscribe-to", "Subscribe to event type full name (e.g. MyNamespace.MyMessage or CustomRuleName,MyNamespace.MyMessage) ", CommandOptionType.MultipleValue);
```
SeanFeldman: There's no mentioning that multiple events are supported

src/CommandLine/Endpoint.cs
danielmarbach: Addressed

danielmarbach:
@SeanFeldman @bording I switched to Brandon's suggestion. That also addresses the review comments from Sean.

danielmarbach:
Update the doco to reflect those changes

src/CommandLine/Endpoint.cs
```diff
@@ -36,5 +36,29 @@ public static async Task Create(ManagementClient client, CommandArgument name, C
...
+            }
+        }
+
+        public static async Task Unsubscribe(ManagementClient client, CommandArgument name, CommandOption topicName, CommandOption subscriptionName, CommandArgument eventType, CommandOption ruleName)
+        {
+            try
+            {
+                await Rule.Delete(client, name, topicName, subscriptionName, eventType, ruleName);
+            }
+            catch (MessagingEntityAlreadyExistsException)
```
bording: Is this the right exception to to be catching here? I think it needs to be `MessagingEntityNotFoundException`, but it would be worth verifying (add a test?)

danielmarbach: I think you're right.

danielmarbach: Fixed

PR #69, opened 2019-07-18
Do not raise critical error for MessageLockLostException

### LINKED ISSUE #70: MessageLockLostException should not trigger critical errors
## Who's affected

Anyone using transport.

## Symptoms

`MessageLockLostException` thrown by the transport is raising a [critical error](https://docs.particular.net/nservicebus/hosting/critical-errors), which in turn can cause endpoints to shutdown.

## Affected supported versions

- 1.2.x
- 1.1.x

## Original issue

**Describe the bug**
When a `MessageLockLostException` occurs during message processing, the error handling pipeline will be invoked. However, the error handling pipeline will also throw the same exception because it can't recover from this exception. When `onError` throws, this error is handled as a critical error.

Some hosts will shut down on critical errors (e.g. SC) which shouldn't happen on a MessageLockLostException.

**To Reproduce**
See this transport test reproducing the issue: https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/69

**Expected behavior**
Not sure whether `onError` should be called in the first place (e.g. moving message to the error queue) as I'd assume the message will be reprocessed anyway. So my first impression is that even the error handling pipeline shouldn't be invoked.

**Versions:**
latest master

timbussmann:
Fixes #70

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -243,12 +243,17 @@
                         await receiver.SafeAbandonAsync(pushSettings.RequiredTransactionMode, lockToken).ConfigureAwait(false);
                     }
                 }
-                catch (Exception onErrorException)
+                catch (Exception onErrorException) when (ExceptionShouldTriggerCriticalError(onErrorException))
                 {
                     criticalError.Raise($"Failed to execute recoverability policy for message with native ID: `{message.MessageId}`", onErrorException);
 
                     await receiver.SafeAbandonAsync(pushSettings.RequiredTransactionMode, lockToken).ConfigureAwait(false);
                 }
+            }
+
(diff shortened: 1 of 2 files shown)


SeanFeldman:
Replaced with #72

src/Transport/Receiving/MessagePump.cs
```diff
@@ -243,13 +243,18 @@ async Task ProcessMessage(Task<Message> receiveTask)
                         await receiver.SafeAbandonAsync(pushSettings.RequiredTransactionMode, lockToken).ConfigureAwait(false);
                     }
                 }
-                catch (Exception onErrorException)
+                catch (Exception onErrorException) when (ExceptionShouldTriggerCriticalError(onErrorException))
```
timbussmann: what's going to happen with the message in case this catch block (which calls `SafeAbandonAsync`) isn't executed?

PR #71, opened 2019-08-14
Bump SDK to latest

danielmarbach:
Just curious what breaks

**Update**
- Updated to new major 4.0.0
- Updated to new minor 4.1.0 with bug fixes
- Updated to new patch 4.1.1 with bug fixes


SeanFeldman:
In theory nothing should. Saying that, we will have to apply a few changes once DevExp or whoever gives ~~us~~ the TF green light.

jbogard:
Any word on this? The new webjobs from MS release also uses 4.x SDK

PR #72, opened 2019-08-22
Do not raise critical error for MessageLockLostException

### LINKED ISSUE #70: MessageLockLostException should not trigger critical errors

SeanFeldman:
Fixes #70

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -243,6 +243,10 @@
                         await receiver.SafeAbandonAsync(pushSettings.RequiredTransactionMode, lockToken).ConfigureAwait(false);
                     }
                 }
+                catch (Exception onErrorException) when (onErrorException is MessageLockLostException || onErrorException is ServiceBusTimeoutException)
+                {
+                    logger.Debug("Failed to execute recoverability.", onErrorException);
+                }
                 catch (Exception onErrorException)
                 {
                     criticalError.Raise($"Failed to execute recoverability policy for message with native ID: `{message.MessageId}`", onErrorException);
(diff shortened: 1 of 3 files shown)


src/Transport/Receiving/MessagePump.cs
```diff
@@ -243,6 +243,10 @@ async Task ProcessMessage(Task<Message> receiveTask)
                         await receiver.SafeAbandonAsync(pushSettings.RequiredTransactionMode, lockToken).ConfigureAwait(false);
                     }
                 }
+                catch (Exception onErrorException) when (onErrorException is MessageLockLostException || onErrorException is ServiceBusTimeoutException)
+                {
+                    logger.Debug("Failed to execute recoverability.", onErrorException);
```
timbussmann: what happens with the message here given that this path won't call `SafeAbandonAsync`?

SeanFeldman: The (reference) to the message is useless to us at this point as we no longer have a valid lock or there's a timeout error, indicating we cannot communicate. The message on the broker is already consumed by another thread/consumer at this point.

timbussmann: 👍 makes sense, thanks for clarifying this. Might have been a useful comment in the code for future readers.

PR #79, opened 2019-10-23
Fix version number by ignoring a faulty commit

SzymonPobiega:


SzymonPobiega:
Version fix seems to be correct as the version on the PR build is correct.

timbussmann:
seems like this test is reliably failing

SeanFeldman:
Strange, no tests were failing. I will rerun it again.

SeanFeldman:
A random single test is failing and looks like it's related to the backend. Specifically, entity management.
Could be a temporary issue with the service.

PR #85, opened 2019-12-01
[SPIKE] Propogate LockedUntiUtc

SeanFeldman:
The idea is to allow the cancellation of an incoming message that cannot be processed due to expired lock.
Message processing can be cancelled using a custom behaviour and `TransportReceiveContext.AbortReceiveOperation` API.


SeanFeldman:
Superceeded by #87

PR #87, opened 2019-12-02
Expose native Message

SeanFeldman:
_Fixes #86_

_Replaces https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/85_

Propagate `Microsoft.Azure.ServiceBus.Message` via message's `ContextBag` to provide the information set by the broker per message.

An example is to allow the cancellation of an incoming message that cannot be processed due to expired lock. The system property, `LockedUntilUtc` could help to decide wherever a message should be attempted for processing or not in case it spent time longer than `MaxLockDuration` while in prefetch. 

With this change, the decision wherever a message should or should not be processed can be made at any point where context is available. E.g. a custom behaviour and `TransportReceiveContext.AbortReceiveOperation()` API, message handler, etc.

### DIFF
--- a/src/AcceptanceTests/Receiving/When_receiving_a_message.cs
+++ b/src/AcceptanceTests/Receiving/When_receiving_a_message.cs
@@ -10,18 +10,18 @@
     public class When_receiving_a_message
     {
         [Test]
-        public async Task Should_have_access_to_the_original_brokered_message_via_the_context_bag()
+        public async Task Should_have_access_to_the_native_message_via_extensions()
         {
             await Scenario.Define<Context>()
                 .WithEndpoint<Endpoint>(b => b.When(
                     (session, c) => session.SendLocal(new Message())))
-                .Done(c => c.OriginalMessageFound)
+                .Done(c => c.NativeMessageFound)
                 .Run();
         }
 
         public class Context : ScenarioContext
         {
-            public bool OriginalMessageFound { get; set; }
+            public bool NativeMessageFound { get; set; }
         }
 
         public class Endpoint : EndpointConfigurationBuilder
@@ -29,7 +29,7 @@
             public Endpoint()
             {
                 EndpointSetup<DefaultServer>((c, d) =>
-                    c.Pipeline.Register(b => new CheckContextForValidUntilUtc(d.ScenarioContext as Context), "Behavior to validate context bag contains the original brokered message"));
+                    c.Pipeline.Register(b => new CheckContextForValidUntilUtc((Context)d.ScenarioContext), "Behavior to validate context bag contains the original brokered message"));
             }
 
             public class Handler : IHandleMessages<Message>
@@ -38,7 +38,7 @@
 
                 public Task Handle(Message request, IMessageHandlerContext context)
                 {
-                    TestContext.OriginalMessageFound = TestContext.OriginalMessageFound && context.Extensions.Get<Microsoft.Azure.ServiceBus.Message>() != null;
+                    TestContext.NativeMessageFound = TestContext.NativeMessageFound && context.Extensions.Get<Microsoft.Azure.ServiceBus.Message>() != null;
 
                     return Task.CompletedTask;
                 }
@@ -55,7 +55,7 @@
 
                 public override Task Invoke(ITransportReceiveContext context, Func<Task> next)
                 {
-                    testContext.OriginalMessageFound = context.Extensions.Get<Microsoft.Azure.ServiceBus.Message>() != null;
+                    testContext.NativeMessageFound = context.Extensions.Get<Microsoft.Azure.ServiceBus.Message>() != null;
 
                     return next();
                 }
--- a/src/AcceptanceTests/Receiving/When_receiving_a_message.cs
+++ b/src/AcceptanceTests/Receiving/When_receiving_a_message.cs
@@ -29,7 +29,8 @@
         {
             public Endpoint()
             {
-                EndpointSetup<DefaultServer>(c => c.Pipeline.Register(b => new CheckContextForValidUntilUtc(b.Build<Context>()), "Behavior to validate context bag contains the original brokered message"));
+                EndpointSetup<DefaultServer>((c, d) =>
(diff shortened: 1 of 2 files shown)


src/AcceptanceTests/Receiving/When_receiving_a_message.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            public DateTime LockedUntilUtcFromHandler { get; set; }
+            public DateTime LockedUntilUtcFromBehavior { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(c => c.Pipeline.Register(b => new CheckContextForValidUntilUtc(b.Build<Context>()), "Behavior to validate context bag contains brokered message system properties"));
```
danielmarbach: There is an overload that allows to pass in the context directly then you don't need to do ` b.Build<Context>`  and can directly register the behavior with newing it up

src/AcceptanceTests/Receiving/When_receiving_a_message.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            public class Handler : IHandleMessages<Message>
+            {
+                public Context TestContext { get; set; }
+
+                public Task Handle(Message request, IMessageHandlerContext context)
+                {
+                    TestContext.LockedUntilUtcFromHandler = context.Extensions.Get<Microsoft.Azure.ServiceBus.Message.SystemPropertiesCollection>().LockedUntilUtc;
```
danielmarbach: I wasn't aware that this is a nested class. I checked now again the ` Message` type and I'm thinking we could really just expose the native message directly if we wanted to. Yes you could then technically do dump stuff with the body but it seems to unlock quite a few interesting native scenarios people can build by having the full power of message available. I remember talking to Brandon and Andreas when they built the ` BasicDeliveryEventArgs`  exposure in RabbitMQ and the conclusion there was that while users can do dangerous stuff it is an advanced low level seam that requires a good understanding of the rabbitmq SDK anyway. I see how we could make the same tradeoffs here

src/AcceptanceTests/Receiving/When_receiving_a_message.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    public class When_receiving_a_message
+    {
+        [Test]
+        public async Task Should_have_access_to_the_original_brokered_message_via_the_context_bag()
+        {
+            await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(b => b.When(
+                    (session, c) => session.SendLocal(new Message())))
+                .Done(c => c.LockedUntilUtcFromHandler == c.LockedUntilUtcFromBehavior)
```
bording: Should the test still be looking this property when we just care about the message type being there?

src/AcceptanceTests/Receiving/When_receiving_a_message.cs
SeanFeldman: That's great, thank you.

src/AcceptanceTests/Receiving/When_receiving_a_message.cs
SeanFeldman: Agreed. Updated.

src/AcceptanceTests/Receiving/When_receiving_a_message.cs
```diff
@@ -0,0 +1,67 @@
...
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+    using Pipeline;
+
+    public class When_receiving_a_message
+    {
+        [Test]
+        public async Task Should_have_access_to_the_original_brokered_message_via_the_context_bag()
```
danielmarbach:
```suggestion
        public async Task Should_have_access_to_the_native_message_via_extensions()
```

src/AcceptanceTests/Receiving/When_receiving_a_message.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            public bool OriginalMessageFound { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>((c, d) =>
+                    c.Pipeline.Register(b => new CheckContextForValidUntilUtc(d.ScenarioContext as Context), "Behavior to validate context bag contains the original brokered message"));
```
danielmarbach: I'd hard cast here

src/AcceptanceTests/Receiving/When_receiving_a_message.cs
```diff
@@ -0,0 +1,67 @@
...
+            await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(b => b.When(
+                    (session, c) => session.SendLocal(new Message())))
+                .Done(c => c.OriginalMessageFound)
+                .Run();
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool OriginalMessageFound { get; set; }
```
danielmarbach: How about `NativeMessageFound` ?

src/AcceptanceTests/Receiving/When_receiving_a_message.cs
```diff
@@ -0,0 +1,67 @@
...
+                readonly Context testContext;
+
+                public CheckContextForValidUntilUtc(Context context)
+                {
+                    testContext = context;
+                }
+
+                public override Task Invoke(ITransportReceiveContext context, Func<Task> next)
+                {
+                    testContext.OriginalMessageFound = context.Extensions.Get<Microsoft.Azure.ServiceBus.Message>() != null;
```
danielmarbach: Should this use ` GetOrDefault` ? if the message is not here this would throw an exception and then cause the test to timeout right?

SeanFeldman: There's GetOrCreate(), but I'd rather keep it as Get. If the message is not found, the test will fail.

PR #178, opened 2020-10-06
CommitableTransaction instead of TransactionScope and supression in the pipeline

danielmarbach:
This PR shows how we could avoid using a behavior in the pipeline by using a `CommitableTransaction` added to the `TransportTransaction`

## Assumptions

It was never supported to wrap a message session with a transaction scope to achieve batched dispatch because message session sends would be isolated and the condition 

```
var shouldSuppressTransaction = !(receiverConnectionAndPathFound && partitionKeyFound) && Transaction.Current != null;
```

prevented that from happening because `receiverConnectionAndPathFound` and `partitionKeyFound` would only be true for received messages under `SendsWithAtomicReceive`. My conclusion was that for all sends we want the following behavior:

- Never use a customer provided TxScope
- For isolated sends never use a TxScope
- For non-isolated sends always use the transports scope if present


src/Transport/Receiving/MessagePump.cs
```diff
@@ -203,24 +203,24 @@ async Task InnerProcessMessage(Task<Message> receiveTask)
...
+                    {
+                        await receiver.SafeCompleteAsync(pushSettings.RequiredTransactionMode, lockToken, transaction)
+                            .ConfigureAwait(false);
 
-                            scope?.Complete();
-                        }
+                        transaction?.Commit();
                     }
 
                     if (receiveCancellationTokenSource.IsCancellationRequested)
```
danielmarbach, line 226: Apparently, when the user requested a cancellation the behavior was always to make the message immediately reappear, therefore abandoning the current message has to happen outside the transaction. I wonder if an explicit Rollback would be good here or is it implicitly clear because the transaction is disposed without ever having called `Commit` (that is how the previous code worked too).

lailabougria: I'd vote for an explicit Rollback personally

danielmarbach:
This needs a few improvements because [`Transaction.Current` doesn't respect `TransactionScopeAsyncFlowOption.Enabled`](https://github.com/dotnet/runtime/issues/29444) and thus stores the current transaction in the thread-local storage.

> Bottom line - the Transaction.Current setter retained its existing behavior of using TLS when TransactionScopeAsyncFlowOption was introduced. In general, mixing TransactionScope and Transaction.Current setter is not a good idea and can also be problematic if Transation.Current is set to a value that does not match what the TransactionScope expects when the TransactionScope is disposed.

This can be worked around by actually wrapping a proper scope with flow option enabled that points to the CommittableTransaction object. 

image

https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-3.0/ms172152(v=vs.85)#completing-a-transaction-scope

I will fix this soon-ish (this was just a first shot to explore if it would work)

danielmarbach:
Did also do a smoke test with the https://docs.particular.net/samples/azure-service-bus-netstandard/send-reply/ and it works like a charm. I used that specific sample because it does a reply as part of handling a message to verify if sends with atomic receive properly works. 

Messages go out only when the receive TX is committed and when I add

```
    class ThrowingBehavior : Behavior<IBatchDispatchContext>
    {
        public override async Task Invoke(IBatchDispatchContext context, Func<Task> next)
        {
            await next().ConfigureAwait(false);

            throw new InvalidOperationException();
        }
    }
```

to the receiving endpoint, no messages go out. I guess that proves that this approach works unless someone finds another loopwhole I haven't covered?

src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -42,18 +42,20 @@ public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction
...
                     var message = transportOperation.Message.ToAzureServiceBusMessage(transportOperation.DeliveryConstraints, partitionKey);
 
-                    using (var scope = CreateTransactionScope(transportOperation.RequiredDispatchConsistency, shouldSuppressTransaction))
+                    using (var scope = transactionToUse.ToScope())
                     {
                         // Invoke sender and immediately return it back to the pool w/o awaiting for completion
                         tasks.Add(sender.SendAsync(message));
-                        scope?.Complete();
+
+                        scope.Complete();
```
danielmarbach, line 58: Would you fancy a comment here that explains the committable tx will not be committed because this scope is not the owner?

src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -65,18 +67,20 @@ public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction
...
```
danielmarbach, line 83: Would you fancy a comment here that explains the committable tx will not be committed because this scope is not the owner?

SzymonPobiega (approved):

SzymonPobiega:
I approve:) Ensures that the transport will work without any problems with NServiceBus.Raw

andreasohlund:
@danielmarbach et al do you know there are any tests around this? (as part of updating to the new Azure ServiceBus SDK we accidentally removed the code that creates the `CommittableTransaction` and all tests where still green)

PR #193, opened 2020-10-30
Do not override disable immediate retries

ramonsmits:


SeanFeldman:
@ramonsmits, what do you plan to do with this PR?

PR #203, opened 2020-11-17
Allow outgoing message customizations

### LINKED ISSUE #202: Customize native outgoing message
We are asking NSB team to provide access to outgoing native ASB message via pipeline just like described here but for incoming messages: https://docs.particular.net/transports/azure-service-bus/native-message-access

SeanFeldman:
Fixes #202 

Scenarios:
- [x] Send using `IMessageSession`
- [x] Publish using `IMessageSession`
- [x] Send using `IMessageHandlerContext`
- [x] Publish using `IMessageHandlerContext`
- [x] Expand support to `IPipelineContext` to allow customizations from behaviors (https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/203/commits/ae7b874f21fef1b2561e471c3c534ad77fc95204)

### DIFF
--- a/src/Transport/Sending/CustomizeNativeMessageExtensions.cs
+++ b/src/Transport/Sending/CustomizeNativeMessageExtensions.cs
@@ -37,10 +37,13 @@
         }
 
         /// <summary>
-        /// Allows customization of the outgoing native message sent using <see cref="IMessageHandlerContext"/>.
+        /// Allows customization of the outgoing native message.
         /// </summary>
+        /// <remarks>
+        /// Messages can be sent using <see cref="IPipelineContext"/> or any of its derived variants such as <see cref="IMessageHandlerContext"/>.
+        /// </remarks>
         /// <param name="options">Option being extended.</param>
-        /// <param name="context"><see cref="IMessageHandlerContext"/> used to dispatch messages in the message handler.</param>
+        /// <param name="context">Context used to dispatch messages in the message handler.</param>
         /// <param name="customization">Customization action.</param>
         public static void CustomizeNativeMessage(this ExtendableOptions options, IPipelineContext context, Action<Message> customization)
         {
--- a/src/Transport/Sending/CustomizeNativeMessageExtensions.cs
+++ b/src/Transport/Sending/CustomizeNativeMessageExtensions.cs
@@ -42,7 +42,7 @@
         /// <param name="options">Option being extended.</param>
         /// <param name="context"><see cref="IMessageHandlerContext"/> used to dispatch messages in the message handler.</param>
         /// <param name="customization">Customization action.</param>
-        public static void CustomizeNativeMessage(this ExtendableOptions options, IMessageHandlerContext context, Action<Message> customization)
+        public static void CustomizeNativeMessage(this ExtendableOptions options, IPipelineContext context, Action<Message> customization)
         {
             if (options.GetHeaders().ContainsKey(CustomizationHeader))
             {
--- a/src/Transport/Sending/CustomizeNativeMessageExtensions.cs
+++ b/src/Transport/Sending/CustomizeNativeMessageExtensions.cs
@@ -15,7 +15,7 @@
         internal const string CustomizationHeader = "$ASB.CustomizationId";
 
         /// <summary>
-        /// Allows customization of the outgoing native message.
+        /// Allows customization of the outgoing native message sent using <see cref="IMessageSession"/>.
         /// </summary>
         /// <param name="options">Option being extended.</param>
         /// <param name="customization">Customization action.</param>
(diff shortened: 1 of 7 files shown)


src/Transport/Sending/CustomizeNativeMessageExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Allows customization of the outgoing native message sent using <see cref="IMessageHandlerContext"/>.
+        /// </summary>
+        /// <param name="options">Option being extended.</param>
+        /// <param name="context"><see cref="IMessageHandlerContext"/> used to dispatch messages in the message handler.</param>
+        /// <param name="customization">Customization action.</param>
+        public static void CustomizeNativeMessage(this ExtendableOptions options, IMessageHandlerContext context, Action<Message> customization)
```
danielmarbach, line 45: If we'd use IPipelineContext here I think we could enable this in the whole receive pipeline including behaviours and not just limit it to handlers. Thoughts?

SeanFeldman: I can't see a reason why not to open up where the customization can be done.
Saying that the evidence I have so far indicates it would be done in the context of the incoming message. Which is not available on `IPipelineContext`. The earliest that has _anything_ related to the incoming message would be `IMessageProcessingContext` with the incoming message headers.

danielmarbach: I don't think that is true. All send operations and extensions on the `IPipelineContext` have options and the only requirement we have is that we can attach a header that we later remove and have access to the extension bag. This is all given on pipeline context. Not sure I understand why incoming headers are relevant

SeanFeldman: The outgoing message customization is there to allow manipulation of the native dispatched message in some business context. That context is the incoming message with the information either in the body or the headers. The [original ask](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/202#issuecomment-724921720) for this feature was driven by this requirement. `IPipelineContext` gives you none of those two. So what good it will do other than being able to assign some arbitrary values to the outgoing native message? Hence my comment about `IMessageProcessingContext` which at least has the incoming headers. Don't focus on the mechanical ability to modify the outgoing message, rather _why_ and _what_ that customization is needed to begin with.

danielmarbach: I do agree in general with what you are saying but then you limit the feature to the original use case. What if the user has other needs and wants to customize the native header with other contextual information as part of a behavior pipeline? Then you would actively prevent such use case scenarios to the pipeline stages that expose `IMessageProcessingContext`. I'll send in a spike to show what I mean

danielmarbach: Raised as a draft https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/205

What the spike confirmed is that since we only care about the context because we need to customize things on the incoming path but we are not exposing it back into the lambda anything concrete that would need to be accessed from the context needs to be extracted before the action delegate anyway and then passed into the action delegate as closure. So even accessing incoming headers would be

```
                    var someHeader = context.MessageHeaders["SomeHeader"];
                    var publishOptions = new PublishOptions();
                    publishOptions.CustomizeNativeMessage(context, m => m.Label = someHeader);
```

This confirms my conclusion that there is no reason to really restrict it to a specific context.

Another thought that came up is if it would make sense to have the context overload do something like

```
                    var publishOptions = new PublishOptions();
                    publishOptions.CustomizeNativeMessage(context, (ctx, m) => m.Label = ctx.MessageHeaders["SomeHeader"]);
```

the upside would be that the contextual information can be accessed in the lambda and it clearly differentiates it, even more, from the `CustomizeNativeMessage` that is "context-free". The drawback would be that we need to capture the context and make them live longer than intended which could cause problems down the line since they are not designed to outlive the stages they are supposed to be used in.

SeanFeldman: > publishOptions.CustomizeNativeMessage(context, (ctx, m) => m.Label = ctx.MessageHeaders["SomeHeader"]);

Feels very awkward. Combined with the fact that you’ve outlined being problematic, I don’t think we should consider it.

> What if the user has other needs and wants to customize the native header with other contextual information as part of a behavior pipeline?

This is what I’m trying to avoid, “what-if”s. Building something concrete and changing when there’s a need rather building based on a guess is what I’m in favour of. Saying that, your original comment did not explain what you had in mind. Rather than “use IPipelineContext” the #205 PR is _adding_ an option of using `IPipelineContext` as an additional overload. That would be a good addition.

danielmarbach: The comment about "use pipeline context" was on the method

```
public static void CustomizeNativeMessage(this ExtendableOptions options, IMessageHandlerContext context, Action<Message> customization)
```

I apologize if that was not precise enough. I felt our communication was going sideways because of a misunderstanding the I probably caused. That's why I raised the PR

SeanFeldman: Yeah, wasn't clear but we've sorted it out and 👍 now.

src/Transport/Sending/CustomizeNativeMessageExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            var nativePropertiesCustomizer = options.GetExtensions().GetOrCreate<NativeMessageCustomizer>();
+            if (!nativePropertiesCustomizer.Customizations.TryAdd(customizationId, customization))
+            {
+                throw new Exception("Failed to apply an outgoing message customization");
+            }
+        }
+
+        /// <summary>
+        /// Allows customization of the outgoing native message sent using <see cref="IMessageHandlerContext"/>.
```
danielmarbach, line 40: Damn I missed the doco updates. Approving so that you can merge even if you decide to address it

src/Transport/Sending/CustomizeNativeMessageExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        /// <param name="options">Option being extended.</param>
+        /// <param name="context"><see cref="IMessageHandlerContext"/> used to dispatch messages in the message handler.</param>
```
danielmarbach, line 43: Same here

danielmarbach (approved):

src/Transport/Sending/CustomizeNativeMessageExtensions.cs
SeanFeldman: Wait... you are human?! 😜

PR #215, opened 2020-12-19
Subscription and rule convention API

### LINKED ISSUE #213: Subscription rules generated with CLI are ignored. "Extra" rules created during Startup. Could push up against ASB rule number limitations.
**Describe the issue**

Using the CLI, we've created subscription rules with short names (`-r` option). When we start the endpoint we see extra rules being created. This can become a problem when the number of rules grows to 1,000 and then the Transport will start creating duplicate rules putting us at the 2,000 limit.

**To Reproduce**

For our development environments we use the NSB CLI to generate subscriptions, rules, etc... 

In many cases we have type names >50 characters, which requires us to have a custom, shorter, rule name. Additionally, in order to have a consistent naming scheme we've applied our custom rule name approach to all of our rules. We have this "RuleNameShortner" configured on the Transport that matches these names we've created.

Examples:

```
asb-transport endpoint subscribe EndpointName MyCompanyName.MyProductName.MyFeatureName.MyClassName -r MyClassName -c "<connection-string>"
asb-transport endpoint subscribe EndpointName MyCompanyName.MyProductName.MyFeatureName.Short -r Short -c "<connection-string>"
```

The above example gives us two rules and both of them are named with the name of the class, even though the `Short` class doesn't require a shortner.

The problem pops up when I start my endpoint. The endpoint will end up creating a new rule for me named `MyCompanyName.MyProductName.MyFeatureName.Short`. It would seem that the "Rule Name Shortner" isn't applied on names that don't need to be shortened. This makes sense, however it seems to conflict with ability to use CLI to create rules with custom names.

**Expected behavior**

Ideally, I'd like to be able to create my topology with CLI (or ARM templates) as desired and have the Transport use what was configured and not create extra rules.

**Versions:**
 - NuGet package: 
   - NServiceBus.AzureFunctions.ServiceBus 0.3.0
   - NServiceBus.Transport.AzureSErviceBus 1.5.0
 - OS: Windows (Azure Function)
 - .NET Core 3.1

SeanFeldman:
Fix #213

### DIFF
--- a/src/Transport/obsoletes-v2.cs
+++ b/src/Transport/obsoletes-v2.cs
@@ -1,0 +1,39 @@
+﻿#pragma warning disable 1591
+
+namespace NServiceBus
+{
+    using System;
+
+    public static partial class AzureServiceBusTransportSettingsExtensions
+    {
+        /// <summary>
+        /// Specifies a callback to apply to the subscription name when the endpoint's name is longer than 50 characters.
+        /// </summary>
+        /// <param name="transportExtensions">The transport settings object</param>
+        /// <param name="subscriptionNameShortener">The callback to apply.</param>
+        // TODO: remove when cherry-picked into release-1.x
+        [ObsoleteEx(Message = "Use `SubscriptionNameConvention()` instead.",
+            RemoveInVersion = "3",
+            TreatAsErrorFromVersion = "2")]
+        public static TransportExtensions<AzureServiceBusTransport> SubscriptionNameShortener(this TransportExtensions<AzureServiceBusTransport> transportExtensions, Func<string, string> subscriptionNameShortener)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Specifies a callback to apply to a subscription rule name when a subscribed event's name is longer than 50 characters.
+        /// </summary>
+        /// <param name="transportExtensions">The transport settings object</param>
+        /// <param name="ruleNameShortener">The callback to apply.</param>
+        // TODO: remove when cherry-picked into release-1.x
+        [ObsoleteEx(Message = "Use `RuleNameConvention()` instead.",
+            RemoveInVersion = "3",
+            TreatAsErrorFromVersion = "2")]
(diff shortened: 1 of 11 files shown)


src/Transport/obsoletes-v2.cs
```diff
@@ -0,0 +1,39 @@
...
+        // TODO: remove when cherry-picked into release-1.x
+        [ObsoleteEx(Message = "Use `RuleNameConvention()` instead.",
+            RemoveInVersion = "3",
+            TreatAsErrorFromVersion = "2")]
+        public static TransportExtensions<AzureServiceBusTransport> RuleNameShortener(this TransportExtensions<AzureServiceBusTransport> transportExtensions, Func<string, string> ruleNameShortener)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
```
SeanFeldman, line 37:
```suggestion
    public static partial class AzureServiceBusTransportSettingsExtensions
    {
        /// <summary>
        /// Specifies a callback to apply to the subscription name when the endpoint's name is longer than 50 characters.
        /// </summary>
        /// <param name="transportExtensions">The transport settings object</param>
        /// <param name="subscriptionNameShortener">The callback to apply.</param>
        [ObsoleteEx(Message = "Use `SubscriptionNameConvention()` instead.",
            RemoveInVersion = "3",
            TreatAsErrorFromVersion = "2")]
        public static TransportExtensions<AzureServiceBusTransport> SubscriptionNameShortener(this TransportExtensions<AzureServiceBusTransport> transportExtensions, Func<string, string> subscriptionNameShortener)
        {
            throw new NotImplementedException();
        }

        /// <summary>
        /// Specifies a callback to apply to a subscription rule name when a subscribed event's name is longer than 50 characters.
        /// </summary>
        /// <param name="transportExtensions">The transport settings object</param>
        /// <param name="ruleNameShortener">The callback to apply.</param>
        [ObsoleteEx(Message = "Use `RuleNameConvention()` instead.",
            RemoveInVersion = "3",
            TreatAsErrorFromVersion = "2")]
        public static TransportExtensions<AzureServiceBusTransport> RuleNameShortener(this TransportExtensions<AzureServiceBusTransport> transportExtensions, Func<string, string> ruleNameShortener)
        {
            throw new NotImplementedException();
        }
    }
}
```

bording (approved):

PR #216, opened 2020-12-19
Subscription and rule convention API

SeanFeldman:
Implementation of #215


danielmarbach:
Why does this have to be a minor? Couldn't you release this as a patch?

SeanFeldman:
> Why does this have to be a minor? Couldn't you release this as a patch?

Unfortunately, it cannot as there are new API options. Those that eventually will deprecate shorteners.

SeanFeldman:
Temporarily targetting `master` to fix an issue with `release-1.8` (needs to be recreated w/o `GitVersion.yml` changes).

src/Transport/obsoletes-v2.cs
```diff
@@ -0,0 +1,71 @@
+﻿#pragma warning disable 1591
```
bording, line 1: For the release branch, we don't need the obsoletes file. These should just be in `AzureServiceBusTransportSettingsExtensions`

src/Transport/NServiceBus.Transport.AzureServiceBus.csproj
```diff
@@ -12,6 +12,8 @@
     <PackageReference Include="NServiceBus" Version="[7.0.1, 8.0.0)" />
     <PackageReference Include="Particular.CodeRules" Version="0.2.1" PrivateAssets="All" />
     <PackageReference Include="Particular.Packaging" Version="0.6.2" PrivateAssets="All" />
+    <PackageReference Include="Fody" Version="6.3.0" PrivateAssets="All" />
```
bording, line 15: I'd prefer to see these alphabetized.

src/TransportTests/ConfigureAzureServiceBusTransportInfrastructure.cs
```diff
@@ -19,7 +19,7 @@ public TransportConfigurationResult Configure(SettingsHolder settings, Transport
 
         var transportExtensions = new TransportExtensions<AzureServiceBusTransport>(settings);
 
-        transportExtensions.SubscriptionNameShortener(name =>
+        transportExtensions.SubscriptionNameConvention(name =>
```
bording, line 22: Do we want the hashed values to be used on everything in the tests? Or perhaps we should a length calculation here?

src/Transport/Configuration/AzureServiceBusTransportSettingsExtensions.cs
```diff
@@ -92,55 +92,53 @@ public static TransportExtensions<AzureServiceBusTransport> TimeToWaitBeforeTrig
...
-        /// Specifies a callback to apply to the subscription name when the endpoint's name is longer than 50 characters.
+        /// Specifies a callback to apply to customize the subscription name.
         /// </summary>
-        /// <param name="transportExtensions"></param>
-        /// <param name="subscriptionNameShortener">The callback to apply.</param>
-        /// <returns></returns>
-        public static TransportExtensions<AzureServiceBusTransport> SubscriptionNameShortener(this TransportExtensions<AzureServiceBusTransport> transportExtensions, Func<string, string> subscriptionNameShortener)
+        /// <param name="transportExtensions">The transport settings object</param>
+        /// <param name="subscriptionNameConvention">The callback to apply.</param>
+        public static TransportExtensions<AzureServiceBusTransport> SubscriptionNameConvention(this TransportExtensions<AzureServiceBusTransport> transportExtensions, Func<string, string> subscriptionNameConvention)
```
bording, line 99: Something is off in the name of the new APIs.

What is the name of the actual entity in ASB? It is "Subscription" or "Subscription name"? 

If we're going to use "convention", it really needs to be something like "naming convention", but "SubscriptionNameNamingConvention" sounds redundant.

I would think it would need to be "SubscriptionNamingConvention"

src/Transport/Configuration/AzureServiceBusTransportSettingsExtensions.cs
```diff
@@ -92,55 +92,53 @@ public static TransportExtensions<AzureServiceBusTransport> TimeToWaitBeforeTrig
...
-        /// Specifies a callback to apply to a subscription rule name when a subscribed event's name is longer than 50 characters.
+        /// Specifies a callback to apply to customize a subscription rule name.
         /// </summary>
-        /// <param name="transportExtensions"></param>
-        /// <param name="ruleNameShortener">The callback to apply.</param>
-        /// <returns></returns>
-        public static TransportExtensions<AzureServiceBusTransport> RuleNameShortener(this TransportExtensions<AzureServiceBusTransport> transportExtensions, Func<string, string> ruleNameShortener)
+        /// <param name="transportExtensions">The transport settings object</param>
+        /// <param name="ruleNameConvention">The callback to apply.</param>
+        public static TransportExtensions<AzureServiceBusTransport> RuleNameConvention(this TransportExtensions<AzureServiceBusTransport> transportExtensions, Func<string, string> ruleNameConvention)
```
bording, line 125: Similar, this should be "RuleNamingConvention", but now that seems too generic. Should this actually be "SubscriptionRuleNamingConvention" ?

bording (changes requested):

src/Transport/Configuration/AzureServiceBusTransportSettingsExtensions.cs
SeanFeldman: The entity is called 'subscription'. It has a name.
`SubscriptionNamingConvention` conventions would work. Similar would be for rules.

src/Transport/Configuration/AzureServiceBusTransportSettingsExtensions.cs
SeanFeldman: `SubscriptionRuleNamingConvention` is a bit chatty but logically a rule is always part of a subscription. We're are replacing the `RuleNameShortener()` API but this could be a good change.

src/Transport/NServiceBus.Transport.AzureServiceBus.csproj
SeanFeldman: I'll push back on this :)

src/Transport/obsoletes-v2.cs
SeanFeldman: aye-aye, captain

src/TransportTests/ConfigureAzureServiceBusTransportInfrastructure.cs
SeanFeldman: Good catch. It should check for the length similar to what other places are doing as we used to use the shortening API.

src/Transport/NServiceBus.Transport.AzureServiceBus.csproj
bording: Why? If they are in that ItemGroup, they should be alphabetized. I'd be fine with them being a in a separate group if you really want to separate them in some way, but as long as they stay in that one, sort them.

SeanFeldman: I like to keep related packages together. With Fody going on top and obsolete.fody to the bottom, the sorting will be alphabetical but the connection is not obvious.

bording (approved):

adamralph:
> 
> 
> Implementation of #215

@SeanFeldman @bording should this not point to https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/213?

SeanFeldman:
This PR is an implementation of #215, which was the actual fix for #213.

SeanFeldman:
@adamralph, this and #215 issues were updated to indicate there was a forward porting of the feature requested in #213.

PR #219, opened 2021-01-06
pass missing exception

SimonCropp:
so the stack trace can be logged

https://discuss.particular.net/t/support-for-net-5/2227


SeanFeldman:
Thank you @SimonCropp. It will be handled and the issue will be updated when done.

timbussmann (approved):

timbussmann:
merging this into a local branch so that TC can run the builds

timbussmann:
thanks for the PR @SimonCropp 👍

PR #225, opened 2021-01-12
GitHubSync update - release-1.5

particularbot6:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


src/CommandLine/CommandRunner.cs
```diff
@@ -13,9 +13,9 @@ public static async Task Run(CommandOption connectionString, Func<ManagementClie
 
             var client = new ManagementClient(connectionStringToUse);
 
-            await func(client);
+            await func(client).ConfigureAwait(false);
```
bording, line 16: We need to revert all the changes to the CommandLine project.

bording (changes requested):

src/CommandLine/CommandRunner.cs
SeanFeldman: I thought I dropped that commit but I didn't. Doing it now.

bording (approved):

PR #227, opened 2021-01-12
GitHubSync update - release-1.7

particularbot6:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


src/CommandLine/.editorconfig
```diff
@@ -0,0 +1,4 @@
+[*.cs]
+
+Add .editorconfig to the CLI tool project
```
bording, line 3: Something is off here

bording (changes requested):

src/CommandLine/.editorconfig
SeanFeldman: Big time, fixing

bording (approved):

PR #230, opened 2021-01-13
Drop missed file

SeanFeldman:
Follow-up for #224


src/CommandLine/.editorconfig
```diff
@@ -0,0 +1,4 @@
+[*.cs]
+
+# Justification: Application project
```
bording, line 3: The original suppression file had "Synchronization context is known" as the reason, so I'm thinking something more like:

```suggestion
# Justification: App synchronization contexts don't require ConfigureAwait(false)
```

bording (approved):

PR #253, opened 2021-02-18
Cache namespace permission check results

timbussmann:
The transport calls `NamespacePermissions.CheckPermission()` multiple times, at startup and runtime (e.g. when creating queues, when creating the subscription, and when subscribing/unsubscribing). Given it's always the same check, it seems to be a lot more efficient to check this once and cache the result (what the subscription manager itself already did).

With this change, we can save some HTTP requests at startup time, improving startup performance.

I've used a `Lazy<Task>` for caching. I think it's not necessary to wrap the call in the factory in a `Task.Run` as usually seen in guidance for async-lazy types as this involves IO that is always going to be awaited directly, so fast returns from the factory method are not that important (there is also no parallelism involved at startup) and we can save the effort to unwrap and so on. Thoughts @danielmarbach ?

### DIFF
--- a/src/Transport/Administration/NamespacePermissions.cs
+++ b/src/Transport/Administration/NamespacePermissions.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Transport.AzureServiceBus
 {
     using System;
+    using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Azure.ServiceBus;
     using Microsoft.Azure.ServiceBus.Management;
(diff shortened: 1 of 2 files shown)


SeanFeldman (approved):

danielmarbach:
Quick question, if there is no parallelism involved at startup wouldn't then a simple bool like outlined in 

```
        async Task CheckForManagePermissions()
        {
            if (!verifiedManagePermissions)
            {
                await namespacePermissions.CanManage().ConfigureAwait(false);
                verifiedManagePermissions = true;
            }
        }
```

but done in the `NamespacePermissions` be sufficient instead of Lazy?

danielmarbach (approved):

tmasternak:
Same question as @danielmarbach. Could we even remove the flag for the proposed snippet? If the connection string does not provide enough permissions we throw and otherwise we never do the check again?

timbussmann:
@danielmarbach yes that would work, especially in the happy case, but it wouldn't provide the same behavior in case of insufficient rights. The Lazy will also cache exceptions and avoid multiple requests in those cases. I don't have a strong opinion on this though.

timbussmann:
looking at the current usage of the API, the simpler field check mentioned by @danielmarbach would suffice because any failure would bubble up to the user and prevent the endpoint from starting in almost all cases.

If the user disables queue creation and pub/sub completely though, we wouldn't check at startup time and it would be only checked when calling `SubscribeAll`/`Unsubscribe` which can potentially be called concurrently (although unlikely). In this case, the local field check would potentially cause parallel HTTP requests being sent, which isn't causing major damage. In this case trying to subscribe/unsubscribe again would re-check every single time though.

I don't have a strong opinion. It feels the simple field check is simpler but a bit dependent on the actual usage of the API whereas the Lazy should be safe to use in all scenarios. What is your preference @danielmarbach @tmasternak ?

tmasternak:
@timbussmann I would say let's go with `Lazy` it's already better than what we had previously. And let's open-up an issue to tackle in an enhancement release.

Issue #173, opened 2020-09-24
asb-transport throws exception when queue already exists


RobCloudlr:
**Describe the bug**
Using asb-transport.exe, when creating a queue, the behavior is different to the behavior of creating an endpoint.  

When you create an endpoint it will create a queue as well ofcourse, it will catch any MessagingEntityAlreadyExistsException and go forward with the process.

When creating a queue, it will not catch the MessagingEntityAlreadyExistsException and stop the process.

**To Reproduce**
Call the following script twice. The creating of the endpoint will mention that it already exists and continue, the creating of the queue will fail with an exception the second time.

asb-transport.exe endpoint create Domein.Hr -c $connectionString
asb-transport.exe queue create error -c $connectionString

**Expected behavior**
I expect the same behavior for queues and topics.

**Versions:**
Latest

**Additional context**
The code of the Endpoint creation with an exception wrapper:
https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/fbeec078f0ca33665e1c7de8208242b0b29b25cd/src/CommandLine/Endpoint.cs#L13-L20

The code of the Queue creation (which is also used by the Endpoint):
https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/fbeec078f0ca33665e1c7de8208242b0b29b25cd/src/CommandLine/Queue.cs#L21

Code that could be catching the exception for the creation of the queue:
https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/fbeec078f0ca33665e1c7de8208242b0b29b25cd/src/CommandLine/Program.cs#L119

Extra information provided by Sean Feldman in this topic:
https://discuss.particular.net/t/operational-scripting-for-azure-service-bus-issue-creating-queue/2096/2

I understand the rationale he's talking about, but I'd expect the rationale to then be applied to both queues and topics. Since this is a tool for automation, I'd expect it it to only create the queues and topics when they don't exist, as we're running the script every deployment.

SeanFeldman:
Thank you @RobCloudlr for raising the issue and elaborating on your POV.
I'm going to add this as a candidate for the next release (no confirmed date).

PR #254, opened 2021-02-19
Skip throwing an exception when a queue already exists.

aleksandr-samila:
Closes https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/173.
* were fixed names in the output for some entities
before the change:
was
and after:
image


SeanFeldman (approved):
`Create_queue_when_it_exists_should_throw()` needs to change as we're changing the behaviour?

danielmarbach (approved):

Issue #358, opened 2021-06-25
expose AutoDeleteOnIdle setting


steve-haar:
**Is your feature request related to a problem? Please describe.**
When using azure cloud machines with AzureServiceBus new queues are created whenever a new cloud instance is brought online. When cloud instances are decommissioned the queues remain even though they will never be used again.

**Describe the solution you'd like**
Expose the AutoDeleteOnIdle property in AzureServiceBusSettings so we can set this property on our queues and topics.

**Describe alternatives you've considered**
The best alternative is to create a task/script that will detect and delete unused queues.

**Additional context**
Here is a patch that I think would solve this.

[0001-add-AutoDeleteOnIdle-setting.patch.txt](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/files/6717964/0001-add-AutoDeleteOnIdle-setting.patch.txt)

SeanFeldman:
Hi @steve-haar,

If you want to suggest a change, opening a [PR](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc) is a common way to suggest changes. But before we get into the changes discussion, could you please clarify a few things first?

> When using azure cloud machines with AzureServiceBus new queues are created whenever a new cloud instance is brought online. When cloud instances are decommissioned the queues remain even though they will never be used again.

1. Are you referring to Azure VMs, VMSS, or something else?
2. When you provision an endpoint with "cloud machine", what drives the logical endpoint name? Is it deterministic or not?
3. How  a new cloud instance is brought online? Is that manual, automated, scripted?

Thank you.

steve-haar:
Sure thing. Here is a PR: https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/359.

1. The situation arises for us because we are hosting NServiceBus endpoints in Azure app services and we are also marking the instances as uniquely addressable (to support callbacks). To make the endpoints uniquely addressable we use the azure environment variable `WEBSITE_INSTANCE_ID` as the discriminator.

2. Because we use `WEBSITE_INSTANCE_ID` as the discriminator, the name is not deterministic.

3.  Azure app services will automatically create and destroy instances of the app service behind the scenes. New instances are created with a random name. Whenever an instance is created, it will create a new queue when it comes online. However, when an instance is destroyed, its queue will stick around forever.

SeanFeldman:
[...]

PR #359, opened 2021-06-28
add AutoDeleteOnIdle setting

steve-haar:
exposes AutoDeleteOnIdle setting for queues and topics.

closes https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/358

### DIFF
--- a/src/Transport/AzureServiceBusTransport.cs
+++ b/src/Transport/AzureServiceBusTransport.cs
@@ -111,6 +111,27 @@
         int entityMaximumSize = 5;
 
         /// <summary>
+        /// The <see cref="TimeSpan"/> idle interval after which topics and queues are automatically deleted.
+        /// </summary>
+        /// <remarks>The minimum duration is 5 minutes. Default value is <see cref="TimeSpan.MaxValue"/>.</remarks>
+        public TimeSpan AutoDeleteOnIdle
+        {
+            get => autoDeleteOnIdle;
+            set
+            {
+                // minimum duration is 5 minutes: https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.subscriptiondescription.autodeleteonidle
+                if (value < TimeSpan.FromMinutes(5))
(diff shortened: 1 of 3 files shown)


src/Transport/AzureServiceBusSettings.cs
```diff
@@ -96,6 +96,19 @@ public AzureServiceBusSettings EntityMaximumSize(int maximumSizeInGB)
             return this;
         }
 
+        /// <summary>
+        /// Overrides the default time to auto delete queues and topics when idle.
+        /// </summary>
+        [ObsoleteEx(
```
SeanFeldman, line 102: The setting (if added) shouldn't be obsoleted.

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -110,6 +110,26 @@ public int EntityMaximumSize
...
+        /// <summary>
+        /// The <see cref="TimeSpan"/> idle interval after which topics and queues are automatically deleted.
+        /// </summary>
+        /// <remarks>The minimum duration is 5 minutes. Default value is <see cref="TimeSpan.MaxValue"/>.</remarks>
+        public TimeSpan AutoDeleteOnIdle
+        {
+            get => autoDeleteOnIdle;
+            set
+            {
+                if (value < TimeSpan.FromMinutes(5))
```
SeanFeldman, line 122: Worth adding a comment with a link to the docs about the minimal value of `QueueDescription.AutoDeleteOnIdle`.

src/Transport/AzureServiceBusSettings.cs
steve-haar: removed setting from `AzureServiceBusSettings`

src/Transport/AzureServiceBusTransport.cs
steve-haar: added comment with link to microsoft docs

PR #393, opened 2021-09-20
Update to the new Azure.Messaging.ServiceBus SDK

WojcikMike:


src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -22,7 +22,7 @@ public MessageDispatcher(MessageSenderPool messageSenderPool, string topicName)
         public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction transaction, CancellationToken cancellationToken = default)
         {
             // Assumption: we're not implementing batching as it will be done by ASB client
-            transaction.TryGet<(ServiceBusConnection, string)>(out var receiverConnectionAndPath);
+            transaction.TryGet<(string, string)>(out var receiverConnectionAndPath);
```
SeanFeldman, line 25: `ServiceBusConnection` has ensured the key is the connection. `string`, if stands for a connection string, will lead to a new connection which should be avoided.

src/Transport/Sending/MessageSenderPool.cs
```diff
@@ -4,59 +4,75 @@
...
-            this.retryPolicy = retryPolicy;
+            this.connectionString = connectionString;
+            this.serviceBusClientOptions = serviceBusClientOptions;
+            serviceBusClientOptions.RetryOptions = retryPolicy;
+            this.tokenCredential = tokenCredential;
+            //TODO: delete this
+            //this.retryPolicy = retryPolicy;
 
-            senders = new ConcurrentDictionary<(string, (ServiceBusConnection, string)), ConcurrentQueue<MessageSender>>();
+            senders = new ConcurrentDictionary<(string, (string, string)), ConcurrentQueue<ServiceBusSender>>();
```
SeanFeldman, line 21: same concern related to `ServiceBusConnection`

src/Transport/Sending/MessageSenderPool.cs
```diff
@@ -4,59 +4,75 @@
...
             }
 
-            var connectionToUse = sender.OwnsConnection ? null : sender.ServiceBusConnection;
-            var path = sender.OwnsConnection ? sender.Path : sender.TransferDestinationPath;
-
-            if (senders.TryGetValue((path, (connectionToUse, sender.ViaEntityPath)), out var sendersForDestination))
+            // var connectionToUse = sender.FullyQualifiedNamespace.OwnsConnection ? null : sender.ServiceBusConnection;
+            // var path = sender.OwnsConnection ? sender.Path : sender.TransferDestinationPath;
+            //
+            // if (senders.TryGetValue((path, (connectionToUse, sender.ViaEntityPath)), out var sendersForDestination))
```
SeanFeldman, line 70: Send-via is done differently in Track 2.
To have cross entity transactional operation, `ServiceBusClientOptions` has to be instructed to use send-via.

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -223,21 +224,36 @@ public TimeSpan TimeToWaitBeforeTriggeringCircuitBreaker
...
             set
             {
-                Guard.AgainstNull(nameof(RetryPolicy), value);
+                Guard.AgainstNull(nameof(RetryPolicyOptions), value);
                 retryPolicy = value;
             }
         }
-        RetryPolicy retryPolicy;
+        ServiceBusRetryOptions retryPolicy;
+        //TODO: old API needs to be deprecated
```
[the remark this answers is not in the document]
WojcikMike: Hey @SeanFeldman ;) Thanks for looking at the PR. I will provide the clarification even that this is still a work in progress. I started with commenting this API due to the types missing from Azure.Messaging.ServiceBus. I saw there TokenCredential abstract class that is implemented in Azure.Identity package.

So at the beginning, it seemed that ITokenProvider went away and was replaced by a different approach of passing a token that can be used to create ASBClient.

src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -41,15 +40,15 @@ public static string GetMessageId(this Message message)
...
-        public static byte[] GetBody(this Message message)
-        {
-            if (message.UserProperties.TryGetValue(TransportMessageHeaders.TransportEncoding, out var value) && value.Equals("wcf/byte-array"))
-            {
-                return message.GetBody<byte[]>() ?? Array.Empty<byte>();
-            }
-
-            return message.Body ?? Array.Empty<byte>();
-        }
+        //TODO: Body???
```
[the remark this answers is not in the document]
WojcikMike: Duly noted, carefully threading. When looking at how to implement that code I had trouble finding corresponding calls in the new API. There seems to be only Body property without the GetBody method. I am uncertain if the client library takes care of the casting, or should we be doing some things. For now, I commented it out to make sure that there is a failing test and that I can verify it.

src/Transport/Sending/MessageDispatcher.cs
WojcikMike: Yeah, I made that change to make the code compile. When we discussed the matter combined with the fact how send-via is done, it seems that the whole MessageSenderPool can only be dependant upon destination. Since the send-via is defined on Client level. So far, the first tests seem to prove it. But we will spend more time on it.

src/Transport/Sending/MessageSenderPool.cs
WojcikMike: Got it, and seems to be done.

src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -42,14 +41,16 @@ public static string GetMessageId(this Message message)
             return message.MessageId;
         }
 
-        public static byte[] GetBody(this Message message)
+
+        //TODO: verify if the new SDK supports sending & receiving "wcf/byte-array"
```
andreasohlund, line 45: @SeanFeldman I've verified this by having an old endpoint send a message to this new version and it was received correctly

`transport.BrokeredMessageBodyType(NServiceBus.Transport.AzureServiceBus.SupportedBrokeredMessageBodyTypes.ByteArray);`

Is there anything else that we might need to test?

SeanFeldman: This was the major thing as the legacy ASB could send a serialized byte array rather than stream of bytes @andreasohlund.
Additional thing to keep in mind is that the new ASB SDK does connection pooling, where the previous client didn't do that. But you've already addressed that.

andreasohlund: Thanks @SeanFeldman ❤️

SeanFeldman: np 🙂 

The cross-entity transaction API got way simpler so that only cuts out some code. Can't wait to see this refreshed transport released. And don't forget to update NSB Functions 🙂 I'll be very eager to update when it comes out.

andreasohlund: Yes, the functions is on our radar

WojcikMike:
So it seems that the failing tests could be fixed by this PR: https://github.com/Azure/azure-sdk-for-net/pull/24215

Currently in the version 7.3.0 of Azure.messaging.servicebus the cancellation token is not passed to ReceivingAmqpLink. With the above PR it is, When the new version is released we should be able to simplify the MessagePump while clause

andreasohlund (approved):

src/Transport/Receiving/MessagePump.cs
```diff
@@ -296,36 +298,38 @@ await receiver.SafeCompleteAsync(transportSettings.TransportTransactionMode, loc
...
+                        await receiver.SafeAbandonMessageAsync(message, transportSettings.TransportTransactionMode, cancellationToken: messageProcessingCancellationToken).ConfigureAwait(false);
                     }
                 }
-                catch (Exception onErrorEx) when (onErrorEx.IsCausedBy(messageProcessingCancellationToken))
+                catch (ServiceBusException onErrorEx) when (onErrorEx.Reason == ServiceBusFailureReason.MessageLockLost || onErrorEx.Reason == ServiceBusFailureReason.ServiceTimeout)
                 {
-                    throw;
+                    Logger.Debug("Failed to execute recoverability.", onErrorEx);
+
+                    await receiver.AbandonMessageAsync(message, cancellationToken: messageProcessingCancellationToken).ConfigureAwait(false);
```
timbussmann, line 322: @andreasohlund @WojcikMike do you remember why this line doesn't use `SafeAbandonAsync` like the other places?

andreasohlund: If I remember correctly there is no need for SafeXyz when you don't want to be "atomic with the receive operation" and since this happens when recoverability has failed you would always want the message to go back to the queue no matter what. (@WojcikMike @mauroservienti please confirm)

That said I don't think it would hurt to use the SafeX version here as well for consistency?

WojcikMike: I honestly do not remember, sorry.

PR #404, opened 2021-09-29
Update Core v7 transport version to use the new Azure.Messaging.ServiceBus SDK

mauroservienti:


src/Transport/Receiving/MessageReceiverExtensions.cs
```diff
@@ -1,45 +1,36 @@
...
     static class MessageReceiverExtensions
     {
-        public static async Task SafeCompleteAsync(this MessageReceiver messageReceiver, TransportTransactionMode transportTransactionMode, string lockToken, Transaction committableTransaction = null)
-        {
-            if (transportTransactionMode != TransportTransactionMode.None)
-            {
-                using (var scope = committableTransaction.ToScope())
-                {
-                    await messageReceiver.CompleteAsync(lockToken).ConfigureAwait(false);
+        //TODO: does this need to be used by the pump?
```
mauroservienti, line 9: @bording while upgrading the transport to use the new SDK we wondered what was the reason to use a dedicated transaction scope when completing or abandoning the message.

Currently, we updated the code to NOT use the extension methods in the class, and instead we simply use the current transaction, e.g.: https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/62a43d7ee626234cabc4f36eef51124d6ea8152c/src/Transport/Receiving/MessagePump.cs#L227-L254

All tests are green. However, we're not sure what was the reason for the extension methods. By any chance, do you happen to remember what was it?

For the records `.ToScope()` does the following:

```
return transaction != null
    ? new TransactionScope(transaction, TransactionScopeAsyncFlowOption.Enabled)
    : new TransactionScope(TransactionScopeOption.Suppress, TransactionScopeAsyncFlowOption.Enabled);
```

/cc @SeanFeldman @andreasohlund @WojcikMike

bording: @mauroservienti That code has changed a bit since Sean and I initially wrote it. See #178 for the details, and how it used to look.

The main goal when we initially wrote the code is that the transaction used by the message pump to interact with the send via support in the client can't be the same transaction that gets exposed to the incoming message pipeline. We were originally doing that with a behavior we added to the pipeline that suppressed the ambient transaction.

Looks like Daniel came up with a different way to do that in #178, and that's where the logic you're asking about was introduced.

Given that it seems like that was added for a specific reason, I think it makes sense to study the changes in #178 and ensure that you aren't breaking something by moving away from that.

SeanFeldman: Unfortunately, I have not been a part of the task force working on that change and can't add more than what Brandon has already provided. Sorry folks.

mauroservienti: Thanks, @bording, and @SeanFeldman. I'll have a look at #178

src/CommandLineTests/NServiceBus.Transport.AzureServiceBus.CommandLine.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
andreasohlund, line 4: Why couldn't we run the tests on .net5?

src/Transport/Receiving/MessagePump.cs
```diff
@@ -162,39 +175,35 @@ async Task InnerProcessMessage(Task<Message> receiveTask)
...
+                if (pushSettings.RequiredTransactionMode != TransportTransactionMode.None)
                 {
-                    // nothing we can do about it, message will be retried
+                    try
+                    {
+                        await receiver.DeadLetterMessageAsync(message, deadLetterReason: "Poisoned message", deadLetterErrorDescription: exception.Message).ConfigureAwait(false);
+                    }
+                    catch (Exception)
+                    {
+                        // nothing we can do about it, message will be retried
```
andreasohlund, line 205: Should we at least log an error?

andreasohlund (approved):

PR #410, opened 2021-10-08
Adding support for cancellation token in MessagePump

WojcikMike:


src/Transport/Receiving/MessagePump.cs
```diff
@@ -108,42 +108,46 @@ public void Start(PushRuntimeSettings limitations)
...
             {
-                logger.Debug($"Exception from {nameof(ProcessMessage)}: ", ex);
+                logger.Error("Error receiving messages.", ex);
             }
             finally
             {
                 try
                 {
-                    semaphore.Release();
+                    _ = semaphore.Release();
```
andreasohlund, line 150: why the assignment?

src/Transport/Receiving/MessagePump.cs
```diff
@@ -169,7 +173,7 @@ async Task InnerProcessMessage(Task<ServiceBusReceivedMessage> receiveTask)
             {
                 // Can happen during endpoint shutdown
             }
-            catch (Exception exception)
+            catch (Exception exception) when (!exception.IsCausedBy(messageReceivingCancellationToken))
```
andreasohlund, line 176: In other message pumps we only use this "when" when catching `OperationCancelled` exception?

src/Transport/Receiving/MessagePump.cs
```diff
@@ -204,9 +208,9 @@ async Task InnerProcessMessage(Task<ServiceBusReceivedMessage> receiveTask)
                 {
                     try
                     {
-                        await receiver.DeadLetterMessageAsync(message, deadLetterReason: "Poisoned message", deadLetterErrorDescription: exception.Message).ConfigureAwait(false);
+                        await receiver.DeadLetterMessageAsync(message, deadLetterReason: "Poisoned message", deadLetterErrorDescription: exception.Message, cancellationToken: messageReceivingCancellationToken).ConfigureAwait(false);
                     }
-                    catch (Exception deadLetterException)
+                    catch (Exception deadLetterException) when (!deadLetterException.IsCausedBy(messageReceivingCancellationToken))
```
andreasohlund, line 213: Same as above

src/Transport/Receiving/MessagePump.cs
WojcikMike: On master: it is on Exception: https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/Receiving/MessagePump.cs#L211 

As the IsCausedBy is an extension method on Exception that checks if it is OperationCancelled: https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/ExceptionExtensions.cs#L9

andreasohlund (approved):

PR #439, opened 2021-11-23
setup-azureservicebus-action

DavidBoike:
There are several flaky tests when the suite is run among multiple frameworks, which is why none of the other PRs in this repo currently pass CI. Rather than fight it, just run each framework separately and benefit from 20m CI times instead of 45m.


.github/workflows/ci.yml
```diff
@@ -69,4 +72,5 @@ jobs:
           connection-string-name: AzureServiceBus_ConnectionString
           tag: ASBTransport
       - name: Run tests
-        uses: Particular/run-tests-action@v1.0.0
+        run: dotnet test src --configuration Release --no-build --framework ${{ matrix.framework }} --logger "GitHubActions;report-warnings=false" -m:1
```
bording, line 75: I wonder if we should add the ability to specify a framework to the run tests action? Main benefit I can think of would be to make this look more consistent with other repos.

bording (approved):

PR #440, opened 2021-11-24
Update to core 8.0.0-alpha.1900

andreasohlund:


src/Transport/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -67,5 +69,24 @@ public override async Task Shutdown(CancellationToken cancellationToken = defaul
                 await messageSenderPool.Close(cancellationToken).ConfigureAwait(false);
             }
         }
+
+        public override string ToTransportAddress(QueueAddress address) => TranslateAddress(address);
+
+        public static string TranslateAddress(QueueAddress address)
```
timbussmann, line 75: is there really a need for the static method here? It seems the only usage of it could be use `ToTransportAddress` anyway? Using the static method, you could remove the parameter from the pump again though.

andreasohlund: Its being called from the transport definition and it needs to be static since the infrastructure might not be created yet?

timbussmann: ok makes sense. Might want to use the static method in the message pump as well then and just skip passing it as a parameter then? Or are you building towards removing the static method once the obsoleted method is fully removed?

andreasohlund: > Or are you building towards removing the static method once the obsoleted method is fully removed?

That was the plan yes

timbussmann (approved):

PR #452, opened 2021-12-10
Update to core 7.6

WojcikMike:


WilliamBZA (approved):
Subject to changing from alpha when core package is released

SzymonPobiega (approved):

Issue #454, opened 2021-12-14
Wire compatibility of messages sent by legacy transport broken in 2.0.1 release


boblangley:
**Describe the bug**
While upgrading to the `Azure.Messaging.ServiceBus` SDK the capability to property decode messages using the `wcf/byte-array` encoding was removed, resulting in wire incompatibility with endpoints sending messages using the [legacy Azure Service Bus transport](https://docs.particular.net/transports/azure-service-bus/compatibility).

Example stack trace (using JSON serialization):

```
NServiceBus.MessageDeserializationException: An error occurred while attempting to extract logical messages from incoming physical message 00000000-0000-0000-0000-000000000000 ---> Newtonsoft.Json.JsonReaderException: Unexpected character encountered while parsing value: @. Path '', line 0, position 0.
at Newtonsoft.Json.JsonTextReader.ParseValue()
at NServiceBus.Newtonsoft.Json.JsonMessageSerializer.IsArrayStream(Stream stream)
at NServiceBus.Newtonsoft.Json.JsonMessageSerializer.Deserialize(Stream stream, IList`1 messageTypes)
at NServiceBus.DeserializeMessageConnector.Extract(IncomingMessage physicalMessage) in /_/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs:line 119
at NServiceBus.DeserializeMessageConnector.ExtractWithExceptionHandling(IncomingMessage message) in /_/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs:line 47
--- End of inner exception stack trace ---
at NServiceBus.DeserializeMessageConnector.ExtractWithExceptionHandling(IncomingMessage message) in /_/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs:line 51
at NServiceBus.DeserializeMessageConnector.<Invoke>d__1.MoveNext() in /_/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs:line 31
--- End of stack trace from previous location where exception was thrown ---
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #455, opened 2021-12-15
Fix legacy compatibility

boblangley:
Relates to #454

### DIFF
--- a/src/Transport/Receiving/MessageExtensions.cs
+++ b/src/Transport/Receiving/MessageExtensions.cs
@@ -2,6 +2,8 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Runtime.Serialization;
+    using System.Xml;
     using Azure.Messaging.ServiceBus;
     using Configuration;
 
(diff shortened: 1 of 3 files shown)


tmasternak (approved):

src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -40,5 +42,24 @@ public static string GetMessageId(this ServiceBusReceivedMessage message)
 
             return message.MessageId;
         }
+
+        public static BinaryData GetBody(this ServiceBusReceivedMessage message)
+        {
+            var body = message.Body ?? new BinaryData(Array.Empty<byte>());
+            var memory = body.ToMemory();
+
+            if (!memory.IsEmpty && message.ApplicationProperties.TryGetValue(TransportMessageHeaders.TransportEncoding, out var value) && value.Equals("wcf/byte-array"))
```
danielmarbach, line 51: I think it would have been beneficial to have a comment here about why we are doing this since this code can be removed May 2022 when the transport is out of support right?

boblangley: I don't think the code can ever be removed. While will no longer provide bug fixes/releases for legacy ASB, we will still be committed to wire compatibility with endpoints running older versions of the software.

PR #462, opened 2022-01-19
Obsolete APIs for next major

aleksandr-samila:
Restored obsolete APIs which are being removed

### DIFF
--- a/src/Transport/obsoletes-v3.cs
+++ b/src/Transport/obsoletes-v3.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.Transport.AzureServiceBus
 {
     using System;
-    using Azure.Core;
     using Azure.Messaging.ServiceBus;
     using Extensibility;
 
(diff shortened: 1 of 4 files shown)


HEskandari (approved):

src/Transport/obsoletes-v3.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    using System;
+    using Azure.Core;
+    using Azure.Messaging.ServiceBus;
+    using Extensibility;
+
+    /// <summary>
+    /// Adds access to the Azure Service Bus transport config to the global Transport object.
+    /// </summary>
+    public static partial class AzureServiceBusTransportSettingsExtensions
```
mikeminutillo, line 11: Looking at this again, it might be better to implement these in the originalclass file and pre-obsolete them. That way they will match [all of the other implementations](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/fa0544789f4cb9412234fc81611f1d77c45c46dd/src/Transport/AzureServiceBusTransportSettingsExtensions.cs)

mikeminutillo (commented):
At least for the Transport Settings Extensions I think we should be able to keep the implementations and pre-obsolete them.

HEskandari (approved):

PR #475, opened 2022-02-15
Add `ServiceBusReceiver` to context for lock renewal to work with transaction mode ReceiveOnly and processing durations over 10 minutes

ramonsmits:
Add ServiceBusReceiver to context for lock renewal to work with transaction mode ReceiveOnly.

Currently the [lock renewal sample for ASBS](https://docs.particular.net/samples/azure-service-bus-netstandard/lock-renewal/) only works with transaction mode SendsAtomicWithReceive. 

A customer currently is affected by the following bugs in the ASB client:

- https://github.com/Azure/azure-sdk-for-net/issues/28025
- https://github.com/Azure/azure-sdk-for-net/issues/30310

This is partially fixed by the following PR but does not mitigate the 10 minutes connection idle timeout

- https://github.com/Azure/azure-sdk-for-net/pull/28351

Benefits:

1. Lock renewal possible with `ReceiveOnly`.
2. Reliable lock renewal for processing durations over 10 minutes
3. Ability to dispatch over 100 messages when using lock renewal because of (1).


SeanFeldman:
>Add `ServiceBusClient` to context which allows behaviors to access it for advanced scenarios. For example, the message lock renewal could then also work with `ReceiveOnly`

@ramonsmits, are there any _other_ significant use cases where a `ServiceBusClient` would be needed?
If it's mostly for lock extension in `ReceiveOnly` mode, I would discourage you from making this change as it will be a short-lived one.

ramonsmits:
> @ramonsmits, are there any _other_ significant use cases where a `ServiceBusClient` would be needed? If it's mostly for lock extension in `ReceiveOnly` mode, I would discourage you from making this change as it will be a short-lived one.

Its a "draft" as while was researching the transport in the context of a support case. This will not be merged.

Could you elaborate on the "as it will be a short-lived one."?

WilliamBZA (approved):

PR #505, opened 2022-03-07
Use TrySetResult instead of SetResult

timbussmann:
These tests occasionally fail (see https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/actions/runs/1934833391) because `SetResult` might be called multiple times, making the `onError` throwback behave differently:

1. message arrives and throws an exception in `onMessage`
2. invoke `onError`
3. `onError` calls `SetResult(true)`
4. `onError` throws `ServiceBusException`
5. internal exception handling calls `AbandonMessage`
6. Message is received again
7. exception thrown in `onMessage`
8. `onError is invoked again`
9. `onError` calls `SetResult(true)` and now throws an `InvalidOperationException` instead
10. The message pump exception handling invokes the critical error action since this in not an expected exception.

### DIFF
--- a/src/TransportTests/When_MessageLockLostException_is_thrown_from_on_error.cs
+++ b/src/TransportTests/When_MessageLockLostException_is_thrown_from_on_error.cs
@@ -24,7 +24,7 @@
                 },
                 (_, __) =>
                 {
-                    onErrorCalled.SetResult(true);
+                    onErrorCalled.TrySetResult(true);
                     throw new ServiceBusException("from onError", ServiceBusFailureReason.MessageLockLost);
                 },
                 transactionMode,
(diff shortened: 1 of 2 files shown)


danielmarbach (approved):
While we are at it could we also switch 

https://github.acom/Particular/NServiceBus.Transport.AzureServiceBus/blob/b20990845969d86efea2200ab59f61b93f84dfe5/src/TransportTests/When_MessageLockLostException_is_thrown_from_on_error.cs#L17

and the other task completion source creations to use the base class method?

Should we also change the core transport tests to always play safe and use the Try** methods?

HEskandari (approved):

mikeminutillo (approved):

mikeminutillo:
@danielmarbach do you mean this one? https://github.com/Particular/NServiceBus/blob/master/src%2FNServiceBus.TransportTests%2FNServiceBusTransportTest.cs#L190

danielmarbach:
@mikeminutillo correct

And I also mentioned to @timbussmann while we are at it we [should use `RunContinuationsAsynchronously`](https://github.com/Particular/NServiceBus/blob/master/src%2FNServiceBus.TransportTests%2FNServiceBusTransportTest.cs#L192) and [fix the memory leaks](https://github.com/Particular/NServiceBus/blob/master/src%2FNServiceBus.TransportTests%2FNServiceBusTransportTest.cs#L196)

danielmarbach (approved):

PR #557, opened 2022-06-15
Testing dotnet7-preview on master

awright18:
Testing dotnet7-preview on master


DavidBoike:
Looks like tests pass on net7.0 with flying colors. @awright18 want to try bumping the library itself to 7.0 as well?

PR #585, opened 2022-08-04
Message pump logger stop event not logging exception data for OperationCanceledException

ramonsmits:
OCE is a gracefully handled exception  and should not result in writing stack trace to logs. Writing exception data is confusing as that usually is correlated with an problem


SeanFeldman:
OCE = `OperationCanceledException`.
Suggest using the fully spelled out name in the issue title over the abbreviation. It helps with searching issues. Makes sense @ramonsmits?

WilliamBZA (approved):

SeanFeldman (approved):

PR #639, opened 2022-10-03
Switch from manually processing to using the service bus processor

danielmarbach:
This PR switches from manually receiving messages towards using the `ServiceBusProcessor` primitive. To keep the shutdown scenario aligned with the previous code, we had to make sure to not pass the shutdown token into the `StopProcessingAsync`  method.

The [lock renewal sample](https://docs.particular.net/samples/azure-service-bus-netstandard/lock-renewal) relies on the context bag to fish out the receiver

```csharp
var messageReceiver = context.Extensions.Get<ServiceBusReceiver>();
```

which needs to be switched to

```csharp
var args = context.Extensions.Get<ProcessMessageEventArgs>();
```

instead. It is arguable whether we still want to support this sample after we have added all the necessary settings again for the auto-lock renewal. Yet adding the args to the context bag makes it at least still possible to renew the lock in similar code like suggested by the sample. Unfortunately, there is no guiding users by having obsoletes in there because we directly expose the SDK types over the context bag.


src/Transport/Receiving/MessagePump.cs
```diff
@@ -271,16 +154,80 @@ async Task ReceiveMessage(CancellationToken messageReceivingCancellationToken)
...
+
+            Logger.Warn(message, processErrorEventArgs.Exception);
+            await circuitBreaker.Failure(processErrorEventArgs.Exception, processErrorEventArgs.CancellationToken)
+                .ConfigureAwait(false);
+        }
+
+        public Task ChangeConcurrency(PushRuntimeSettings newLimitations, CancellationToken cancellationToken = default)
+        {
+            limitations = newLimitations;
+            processor.UpdateConcurrency(limitations.MaxConcurrency);
```
danielmarbach, line 185: Unfortunately, it looks like when https://github.com/Particular/NServiceBus/pull/6214 was done the tests have only be added to the Core folder which means they are not executed in downstreams. So all transport repositories are essentially flying blind as of today and we need to manually verify this or copy and paste some of the acceptance tests from Core into this repository even though all other transports have the same issue.

danielmarbach: Previous version did when Starting again adjust the prefetch multipler. I will see if that was just a side effect of the impl choice or whether it was intentionally chosen like that.

I'm happy to roll back otherwise. I would love though to use what the SDK provides.

ramonsmits:
Can't we just enabled lock renewal by default?

danielmarbach:
It is already on by default on the processor :)

Further PRs will expose the settings. This is just the first step

lailabougria (approved):

PR #644, opened 2022-10-04
Batching support for non-isolated dispatches

### LINKED ISSUE #517: Leverage built-in batching
With the new SDK it is possible to more or less elegantly batch messages. TryAddMessage handles the message size restrictions based on the tier of the namespace automatically. It d[oesn't yet handle the maximum allowed messages to be added to a batch based on the tier](https://github.com/Azure/azure-sdk-for-net/issues/21451) but that will eventually be built into the batch as soon as the broker reports that back to the SDK.

```csharp
 var messageCount = messagesToSend.Count;
 int batchCount = 1;
 while (messagesToSend.Count > 0)
 {
     using ServiceBusMessageBatch messageBatch = await sender.CreateMessageBatchAsync();
 
     if (messageBatch.TryAddMessage(messagesToSend.Peek()))
     {
         messagesToSend.Dequeue();
     }
     else
     {
         throw new Exception($"Message {messageCount - messagesToSend.Count} is too large and cannot be sent.");
     }
 
     while (messagesToSend.Count > 0 && messageBatch.TryAddMessage(messagesToSend.Peek()))
     {
         messagesToSend.Dequeue();
     }
 
     await sender.SendMessagesAsync(messageBatch);
 }
```

danielmarbach:
Closes #517 

This pull request introduces (batching support for non-isolated dispatches. It leverages the SDKs batching mechanism that also allows to check for the message size and creating multiple batches in accordance with the SKU. 

`CreateMessageBatchAsync` is an operation that returns a value task. The value task is only there to query the SKU capability once in case it hasn't been done. From that point on the information is "cached" and batch creation is synchronous and in-memory. So there is no impact on the throughput when using batches for single messages. Once multiple messages are sent the messages will be packaged into a single call to the broker (as long as they fit the batch) which overall reduces the exposure to latency of the link between the endpoint and the broker.

This also touches the way we manage senders. The previous code was ported over from the Microsoft.Azure.ServiceBus SDK which did have individual connections per message sender (exception for cross entity transactions, they are required to share the connection). The code assumed it is beneficial to get a sender from the pool, kick off the operation, and then return the sender immediately to the pool. Even though this worked it did rely on the SDK enlistening on the current transaction on the synchronous path of the code. Should this code be refactored those assumptions might break. On top of that when the switch to the new SDK was made the code has not been changed in the pool but the SDK started to behave differently, namely the connection was shared between the client (parent) and the sender. So that means already with the released version the original assumptions of "pooling makes things faster" are no longer true because the code has changed in the meantime.

The new approach makes sure there is one sender per destination / client pair that is used over the lifetime of the endpoint. This is aligned with the best practices of treating senders as singletons and that they can be shared across threads.

Should we wish to introduce multiple connections again like the original pooling mechanism assumed, we would need to introduce anyway a different mechanics of having multiple client/sender pairs for sends outside the handling pipeline or for endpoints running with `None` or `ReceiveOnly` 

More information also see https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/MigrationGuide.md#connection-pooling


lailabougria (approved):

src/Transport/Sending/MessageSenderRegistry.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            this.defaultClient = defaultClient;
+
+            destinationToSenderMapping = new ConcurrentDictionary<(string, ServiceBusClient), Lazy<ServiceBusSender>>();
+        }
+
+        public ServiceBusSender GetMessageSender(string destination, ServiceBusClient client)
+        {
+            // According to the client SDK guidelines we can safely use these client objects for concurrent asynchronous
+            // operations and from multiple threads.
```
danielmarbach, line 22:
```suggestion
            // According to the client SDK guidelines we can safely use these client objects for concurrent asynchronous
            // operations and from multiple threads.
            // see https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-performance-improvements?#reusing-factories-and-clients
```

src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -10,108 +12,205 @@
...
+                    if (Log.IsDebugEnabled)
+                    {
+                        added.ApplicationProperties.TryGetValue(Headers.MessageId, out var messageId);
+                        logBuilder!.Append($"{messageId ?? added.MessageId},");
+                    }
+                }
+                else
+                {
+                    peekedMessage.ApplicationProperties.TryGetValue(Headers.MessageId, out var messageId);
+                    throw new ServiceBusException($"Message #{messageCount - messagesToSend.Count} with message id '{messageId ?? peekedMessage.MessageId}' is too large and cannot be sent.", ServiceBusFailureReason.MessageSizeExceeded);
```
SeanFeldman, line 154: >"Message #N with message id 'XXXXXXXXXXX' is too large and cannot be sent."

That's not necessarily accurate or helpful.

Realistically, thinking about why it could happen, there are 2 common scenarios:
1. Too many messages combined together exceeded the maximum and
2. One of the messages was big.

It would help to throw an exception with messages attempted for dispatch as a batch
- Total size of the 
- Message types involved in the batch

SeanFeldman (commented):
sorry, late to the party

src/Transport/Sending/MessageDispatcher.cs
danielmarbach: @SeanFeldman thanks for the feedback. You are not late to the party. We are planning to do a few more rounds of improvements. I will send in a PR that also addresses your input :crossed_fingers:

danielmarbach: @SeanFeldman I was thinking about this a bit more and I think there is not much we can do without significantly impacting the success path as well as reimplementing parts of the logic that is ultimately part of the batch and should be owned by the SDK. 

Because we are building up the batch by peeking we would need to store all the information that might be helpful to be included in the case when a batch fails proactively into some structure that is only ever used when the `TryAdd` returned `false`. This is because the batch itself doesn't allow to be enumerated over the content it owns. In order to get the total size of the message we would need to do a back of the napkin calculation in this case too. The benefit of the batch as it stands today is that it encapsulates logic of the service bus tier, message length calculations and more which we would have to duplicate up again into the code. 

I will try to see if I can make the exception message slightly more helpful but going down the path of storing a lot of additional contextual information plus re-implementing parts of the SDK for the failure case is not what we would like to do. 

Can you maybe think of an acceptable tradeoff? FYI I'm not rejecting the input. I will continue to think about better approaches.

SeanFeldman: Makes sense. In that case, something that would guide towards troubleshooting w/o leading inna direction that won't be much of a help. Like the message #. Perhaps a big blob of text explaining how this could have happened, to guide the users towards fixing what only they can fix. E.g.
"The messages attempted for dispatch have either collectively exceeded the maximum batch size N or one of the messages has the maximum size of N. Review your code, reduce messages size by using data bus, or use the Premium Service Bus tier." Something amongst those lines.

danielmarbach: What do you think about this https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/649 ?

PR #648, opened 2022-10-22
Client management improvements

danielmarbach:
Additional enhancements as a follow-up to #644 

- Makes sure all senders are properly disposed
- Makes sure all clients including the default client are disposed
- Makes sure the default client has never the cross entity transaction flag set to avoid accidentally using the feature where it makes no sense to use it
- Creates dedicated option types for the receivers and the default clients (for better clarity, even though the option type is internally cloned)
- Assigns more descriptive IDs to the various client types

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -91,7 +91,7 @@
                 ReceiveMode = transportSettings.TransportTransactionMode == TransportTransactionMode.None
                     ? ServiceBusReceiveMode.ReceiveAndDelete
                     : ServiceBusReceiveMode.PeekLock,
-                Identifier = Id,
+                Identifier = $"Processor-{Id}-{ReceiveAddress}-{Guid.NewGuid()}",
                 MaxConcurrentCalls = limitations.MaxConcurrency,
                 AutoCompleteMessages = false
             };
--- a/src/Tests/FakeSender.cs
+++ b/src/Tests/FakeSender.cs
@@ -40,8 +40,7 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests
             return Task.CompletedTask;
         }
 
-        public override Task SendMessagesAsync(ServiceBusMessageBatch messageBatch,
-            CancellationToken cancellationToken = new CancellationToken())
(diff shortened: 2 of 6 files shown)


src/Transport/Receiving/MessagePump.cs
```diff
@@ -91,7 +91,7 @@ public async Task StartReceive(CancellationToken cancellationToken = default)
                 ReceiveMode = transportSettings.TransportTransactionMode == TransportTransactionMode.None
                     ? ServiceBusReceiveMode.ReceiveAndDelete
                     : ServiceBusReceiveMode.PeekLock,
-                Identifier = Id,
+                Identifier = $"Processor-{Id}-{ReceiveAddress}-{RandomIdGenerator.GetBase36(8)}",
```
lailabougria, line 94: Why do we need the random id here?

danielmarbach: By default, the SDK assigns a random ID of type `Guid` to it. Since we took over the control to customize the ID we need to make sure different instances of the clients are uniquely identifiable. With this approach, we have something that is more humane readable and also shorter than a GUID plus we have enough randomness that makes sure when several instances are running on the same node they are uniquely identifiable.

I also thought about adding the endpoint name to all of them, but that would make everything super long, and it would require some advanced trickery in the message registry to avoid the closures so I only added the endpoint name as an identifier for the default instance

danielmarbach: The other option would be that we do something like

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Hosting/HostingComponent.Settings.cs#L103

But we cannot use MD5 like the core does (see the comment there). So we would need to get access to the executable path, the machine name from the settings and then deterministically hash that. Then take the bytes from the GUID that is most likely unique enough. This would keep the ID consistent over restarts (which I doubt we really need here).

danielmarbach: Discussed it with @lailabougria and we concluded traces already have quite a few GUIDs in there, so trying to create more human-readable IDs isn't necessarily a goal we need to achieve. Switching back to GUIDs.

lailabougria (approved):

PR #650, opened 2022-10-24
Expose the MaxAutoLockRenewalDuration

danielmarbach:
Builds on top of #639 

Exposes the `MaxAutoLockRenewalDuration` from the SDK and forwards that to the `ServiceBusProcessor` used in the message pump. The processor will continue to extend the message lock up to the specified time.

The SDK itself uses a default value of 5 minutes. I did only expose that fact over the documentation, but not hard-code the 5 minutes into the code. This makes sure we follow the same defaults as the underlying SDK. Another option would be to remove the remark entirely and just rely on the SDK defaults without mentioning on the documentation.

It is also possible to manually extend the lock in the message handling pipeline by getting access to the processor arguments available on the context (introduced in the referenced PR).

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -96,6 +96,11 @@
                 AutoCompleteMessages = false
             };
 
+            if (transportSettings.MaxAutoLockRenewalDuration.HasValue)
+            {
+                receiveOptions.MaxAutoLockRenewalDuration = transportSettings.MaxAutoLockRenewalDuration.Value;
+            }
+
             processor = serviceBusClient.CreateProcessor(ReceiveAddress, receiveOptions);
             processor.ProcessErrorAsync += OnProcessorError;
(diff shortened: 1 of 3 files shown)


src/Transport/Receiving/MessagePump.cs
```diff
@@ -96,6 +96,11 @@ public async Task StartReceive(CancellationToken cancellationToken = default)
                 AutoCompleteMessages = false
             };
 
+            if (transportSettings.MaxAutoLockRenewalDuration.HasValue)
+            {
+                receiveOptions.MaxAutoLockRenewalDuration = transportSettings.MaxAutoLockRenewalDuration.Value;
```
lailabougria, line 101: Agreed we should follow the SDK's default, but we should document this as it might result in behavior changes when upgrading to new versions of the SDK.

danielmarbach: Documented it in the corresponding internal design document

PR #653, opened 2022-10-25
Enable the command line tool to use a namespace with DefaultAzureCredentials

### LINKED ISSUE #562: Please use DefaultAzureCredential in CommandLine for CI/CD pipelines
Since we switched to managed identity, our ops team wants to use role-based assignments for everything, and are strongly opposed to any use of a SharedAccessKey.

We need to run the CommandLine in a "Azure PowerShell" or "Azure CLI" task in Azure DevOps (i.e. using a registered service connection), but it fails because it only supports connection strings.

I would like to modify https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/a3f004e7c8dfcb9e96981f03ac5af793b4e47413/src/CommandLine/CommandRunner.cs#L12-L14

To be something like:
```csharp
            var connectionStringToUse = connectionString.HasValue() ? connectionString.Value() : Environment.GetEnvironmentVariable(EnvironmentVariableName);
            
            ServiceBusAdministrationClient client;
            if (connectionStringToUse.Contains("="))
            {
                client = new ServiceBusAdministrationClient(connectionStringToUse);
            }
            else
            {
                client = new ServiceBusAdministrationClient(connectionStringToUse, new DefaultAzureCredential());
            }
```

If we use the [token overload](https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.administration.servicebusadministrationclient.-ctor?view=azure-dotnet#azure-messaging-servicebus-administration-servicebusadministrationclient-ctor(system-string-azure-core-tokencredential)) like that, the tool would just pick up the existing PowerShell or Az CLI authentication token (plus handling environment variables, etc.) (See [DefaultAzureCredential](https://docs.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet)).

danielmarbach:
Closes #562 

I have thought about adding tests to verify the credential chain works but ultimately figured it is too complex and not adding much value. Having tests that verify that part of the tool would require either every developer to have additional credentials on the machine or we could conditionally only execute the tests on the CI. For those, the azure/login action would establish the token and we could execute the tests but ultimately, we would only be testing the SDK so I figured it is not worth the complexity.

### DIFF
--- a/src/CommandLine/Program.cs
+++ b/src/CommandLine/Program.cs
@@ -21,7 +21,7 @@
 
             var fullyQualifiedNamespace = new CommandOption("-n|--namespace", CommandOptionType.SingleValue)
             {
-                Description = "Sets the fully qualified namespace for connecting without the connection string"
+                Description = "Sets the fully qualified namespace for connecting with cached credentials, such as those from Azure PowerShell or CLI"
             };
             fullyQualifiedNamespace.OnValidate(v => ValidateConnectionAndNamespaceNotUsedTogether(connectionString, fullyQualifiedNamespace));
             connectionString.OnValidate(v => ValidateConnectionAndNamespaceNotUsedTogether(connectionString, fullyQualifiedNamespace));
--- a/src/CommandLine/Program.cs
+++ b/src/CommandLine/Program.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Transport.AzureServiceBus.CommandLine
 {
     using System;
+    using System.ComponentModel.DataAnnotations;
     using Azure.Messaging.ServiceBus;
     using McMaster.Extensions.CommandLineUtils;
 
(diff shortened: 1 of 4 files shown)


abparticular (approved):

src/CommandLine/Program.cs
```diff
@@ -18,6 +19,13 @@ static int Main(string[] args)
                 Description = $"Overrides environment variable '{CommandRunner.EnvironmentVariableName}'"
             };
 
+            var fullyQualifiedNamespace = new CommandOption("-n|--namespace", CommandOptionType.SingleValue)
+            {
+                Description = "Sets the fully qualified namespace for connecting without the connection string"
```
danielmarbach, line 24:
```suggestion
                Description = "Sets the fully qualified namespace for connecting with cached credentials, such as those from Azure PowerShell or CLI"
```

PR #654, opened 2022-10-25
Some test coverage for the pump

danielmarbach:
Brings in some test coverage for some of the scenarios of the pump. Not everything is covered, and there is still some redundancy that I deliberately did not yet remove. With this in place, we hopefully are in a much better state to continue to evolve the pump and add more tests. For example, it is now possible to verify the processor event args are added to the context (which I will add as a subsequent PR).

### DIFF
--- a/src/Tests/FakeServiceBusClient.cs
+++ b/src/Tests/FakeServiceBusClient.cs
@@ -6,24 +6,36 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests
     public class FakeServiceBusClient : ServiceBusClient
     {
         public Dictionary<string, FakeSender> Senders { get; } = new Dictionary<string, FakeSender>();
+        public Dictionary<string, FakeProcessor> Processors { get; } = new Dictionary<string, FakeProcessor>();
 
         public override ServiceBusSender CreateSender(string queueOrTopicName)
         {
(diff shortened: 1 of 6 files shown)


src/Tests/FakeServiceBusClient.cs
```diff
@@ -25,5 +26,15 @@ public override ServiceBusSender CreateSender(string queueOrTopicName, ServiceBu
...
+
+        public override ServiceBusProcessor CreateProcessor(string queueName, ServiceBusProcessorOptions options)
+        {
+            if (!Processors.ContainsKey(queueName))
+            {
+                var fakeSender = new FakeProcessor();
+                Processors[queueName] = fakeSender;
+            }
+            return Processors[queueName];
+        }
```
abparticular, line 38: I've heard you like performance tips so I'll provide an alternative for your consideration :smile: (not that it will matter in this case)

```
            if (!Processors.TryGetValue(queueName, out var fakeSender))
            {
                fakeSender = new FakeProcessor();
                Processors.Add(queueName, fakeSender);
            }
            return fakeSender;
```

danielmarbach: Adressed :dancing_men:

PR #658, opened 2022-10-28
Make the token credential and the fullyQualifiedNamespace a dedicated connection setting instead of misusing the connection string.

### LINKED ISSUE #604: When using `TokenCredential` validate if connection string only contains the fully qualified namespace
When using `TokenCredential` the transport connection string needs to different:

Wrong:

```c#
transport.ConnectionString("Endpoint=sb://[NAMESPACE].servicebus.windows.net/;SharedAccessKeyName=[KEYNAME];SharedAccessKey=[KEY]");
```
Correct:

```c#
transport.ConnectionString("[NAMESPACE].servicebus.windows.net");
```

Not doing this results in:

> System.ArgumentException: The value 'Endpoint=sb://my-namespace.servicebus.windows.net/' is not a well-formed Service Bus fully qualified namespace. (Parameter 'fullyQualifiedNamespace')

See:

- https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/588

It is not obvious that this is due to an incorrect connection string.

[The guidance has been updated to clarify how to use TokenCredential](https://github.com/Particular/docs.particular.net/pull/5830) but we can do better.

When using `TokenCredential` then 

a. the connection string must be validated to only contains a fully qualified namespace
b. or, the fully qualified namespace should be extracted from the connection string via `ServiceBusConnectionStringProperties.Parse(ConnectionString).FullyQualifiedNamespace`.

danielmarbach:
- Fixes #604

### DIFF
--- a/src/Transport/AzureServiceBusTransportSettingsExtensions.cs
+++ b/src/Transport/AzureServiceBusTransportSettingsExtensions.cs
@@ -233,6 +233,7 @@
         public static TransportExtensions<AzureServiceBusTransport> CustomTokenCredential(this TransportExtensions<AzureServiceBusTransport> transportExtensions, string fullyQualifiedNamespace, TokenCredential tokenCredential)
         {
             Guard.AgainstNull(nameof(tokenCredential), tokenCredential);
+            Guard.AgainstNullAndEmpty(nameof(fullyQualifiedNamespace), fullyQualifiedNamespace);
             transportExtensions.Transport.FullyQualifiedNamespace = fullyQualifiedNamespace;
             transportExtensions.Transport.TokenCredential = tokenCredential;
             return transportExtensions;
--- a/src/Transport/AzureServiceBusTransportSettingsExtensions.cs
+++ b/src/Transport/AzureServiceBusTransportSettingsExtensions.cs
@@ -224,14 +224,16 @@
         /// Overrides the default token credential with a custom one.
         /// </summary>
         /// <param name="transportExtensions"></param>
+        /// <param name="fullyQualifiedNamespace">The fully qualified namespace.</param>
         /// <param name="tokenCredential">The token credential to be used.</param>
         [PreObsolete(
-            ReplacementTypeOrMember = "AzureServiceBusTransport.TokenCredential",
+            ReplacementTypeOrMember = "AzureServiceBusTransport(string fullyQualifiedNamespace, TokenCredential tokenCredential)",
             TreatAsErrorFromVersion = "4",
             RemoveInVersion = "5")]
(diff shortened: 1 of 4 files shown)


src/Transport/obsoletes-v3.cs
```diff
@@ -22,6 +23,12 @@ public static partial class AzureServiceBusTransportSettingsExtensions
             this TransportExtensions<AzureServiceBusTransport> transportExtensions,
             Func<string, string> subscriptionNameShortener)
             => throw new NotImplementedException();
+
+        [ObsoleteEx(ReplacementTypeOrMember = "CustomTokenCredential(string fullyQualifiedNamespace, TokenCredential tokenCredential)",
+            TreatAsErrorFromVersion = "3",
+            RemoveInVersion = "4")]
+        public static TransportExtensions<AzureServiceBusTransport> CustomTokenCredential(this TransportExtensions<AzureServiceBusTransport> transportExtensions, TokenCredential tokenCredential)
```
bording, line 30: Why is this obsolete? Doesn't this mean there's no way to use a `TokenCredential` in the not obsolete `Transport<T>` API?

bording: That's what I get for not pulling this down and looking at it locally. I find the GitHub UI for PR reviews to be rather terrible!

src/Transport/AzureServiceBusTransportSettingsExtensions.cs
```diff
@@ -218,14 +218,16 @@ public static TransportExtensions<AzureServiceBusTransport> CustomRetryPolicy(th
...
         [PreObsolete(
-            ReplacementTypeOrMember = "AzureServiceBusTransport.TokenCredential",
+            ReplacementTypeOrMember = "AzureServiceBusTransport(string fullyQualifiedNamespace, TokenCredential tokenCredential)",
             TreatAsErrorFromVersion = "4",
             RemoveInVersion = "5")]
-        public static TransportExtensions<AzureServiceBusTransport> CustomTokenCredential(this TransportExtensions<AzureServiceBusTransport> transportExtensions, TokenCredential tokenCredential)
+        public static TransportExtensions<AzureServiceBusTransport> CustomTokenCredential(this TransportExtensions<AzureServiceBusTransport> transportExtensions, string fullyQualifiedNamespace, TokenCredential tokenCredential)
         {
             Guard.AgainstNull(nameof(tokenCredential), tokenCredential);
+            transportExtensions.Transport.FullyQualifiedNamespace = fullyQualifiedNamespace;
```
bording, line 230:
```suggestion
            Guard.AgainstNullAndEmpty(nameof(fullyQualifiedNamespace), fullyQualifiedNamespace);
            transportExtensions.Transport.FullyQualifiedNamespace = fullyQualifiedNamespace;
```

bording (changes requested):

bording (approved):

PR #660, opened 2022-10-28
Skip executing the pipeline for messages that have already lost the lock

danielmarbach:
Unfortunately, I couldn't find a way to avoid the potential clock drift error. 

I have also raised [a feature request into the SDK](https://github.com/Azure/azure-sdk-for-net/issues/31966)

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -127,6 +127,16 @@
             try
             {
                 messageId = message.GetMessageId();
+
+                if (processor.ReceiveMode == ServiceBusReceiveMode.PeekLock && message.LockedUntil < DateTimeOffset.UtcNow)
+                {
+                    if (Logger.IsDebugEnabled)
+                    {
+                        Logger.Debug($"Skip handling the message with id '{messageId}' because the lock has expired at '{message.LockedUntil}'.");
+                    }
+                    return;
+                }
+
                 headers = message.GetNServiceBusHeaders();
                 body = message.GetBody();
             }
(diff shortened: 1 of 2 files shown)


ramonsmits (approved):
This would already resolve issues with already expired prefetched messages.

src/Transport/Receiving/MessagePump.cs
```diff
@@ -127,6 +127,16 @@ async Task OnProcessMessage(ProcessMessageEventArgs arg)
             try
             {
                 messageId = message.GetMessageId();
+
+                if (processor.ReceiveMode == ServiceBusReceiveMode.PeekLock && message.LockedUntil < DateTimeOffset.UtcNow)
+                {
+                    if (Logger.IsDebugEnabled)
+                    {
+                        Logger.Debug($"Skip handling the message with id '{messageId}' because the lock has expired at '{message.LockedUntil}'.");
```
SeanFeldman, line 135: A message is losing its lock token primarily due to the misconfigured consumer. Either not enough processing "power," or the processing takes way too long. In this case, the message would start processing after waiting too long in the in-memory queue (prefetched) and lose its lock token before even getting processed by the handler, which is indicative of an endpoint that needs to be reconfigured (less aggressive prefetch, to be scaled out to handle the load, etc.). As such, this is not a debug-level message that should be logged but rather a **warning** to provide monitoring with information about potential misconfiguration/scaling issues. 

While debug level is better than nothing, there's a high probability that the issue is not going to be just resolved and messages will be skipped multiple times. Combined with `DeliveryCount` set to a very high number, this would cause unnecessary strain on the endpoint and excessive charges. Another reason why it should be treated as a warning.

SeanFeldman (commented):
Did I already mention I'm always late to the party? 🙂

src/Transport/Receiving/MessagePump.cs
danielmarbach: Thanks @SeanFeldman. As always your input is very valuable. Will fix

danielmarbach: Done

SeanFeldman (approved):

abparticular (approved):

PR #662, opened 2022-10-31
Polyfill the MaxAutoLockRenewalDuration

danielmarbach:
Follow up for #650 to make sure the old API supports setting this value too.


src/Transport/AzureServiceBusTransportSettingsExtensions.cs
```diff
@@ -11,14 +11,17 @@
     /// </summary>
     public static partial class AzureServiceBusTransportSettingsExtensions
     {
+        const string Note = "As long as the persistence configuration API has not been adjusted to match the transport configuration API keep bumping the versions when working on a new major";
```
abparticular, line 14: Is there an issue or documentation we can link to help Future Dev determine if the persistence configuration has been adjusted to match the transport configuration?

danielmarbach: Unfortunately, at this stage we have only an internal tracking issue that we cannot link in the public documentation :(

danielmarbach:
Apparently, it is not entirely clear whether we should or should not polyfill. I will be extracting the notes changes into a dedicated PR that we can merge independently of the shim API changes

danielmarbach:
https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/664

lailabougria (approved):
Given that in this PR, we're going the extra mile for the user experience to provide options for both flavors of the API. Given that there's no full agreement and the action points item regarding this topic, I think we can merge this as is. If the conclusion is that we shouldn't go the extra mile, we can still remove it later.

PR #669, opened 2022-11-03
Honor the maximum batch size enforced by the service

danielmarbach:
When `SendsAtomicReceive` is used it is not possible to send more than 100 messages in a single transaction. 

> 2022-11-03 15:44:08.047 ERROR Exception from Send.
> Azure.Messaging.ServiceBus.ServiceBusException: Cannot send more than 100 messages in a single transaction. 

Unfortunately, `TryAdd` might still return true because `TryAdd` is not aware of this restriction. Only sending the messages with a transaction scope will enforce this restriction. 

There is another limit at play for non-transactional sends. That magic limit of 4500 messages is depending on the tier and should be incorporated at some point into the [`TryAdd` method](https://github.com/Azure/azure-sdk-for-net/issues/21451).

### DIFF
--- a/src/Tests/Sending/MessageDispatcherTests.cs
+++ b/src/Tests/Sending/MessageDispatcherTests.cs
@@ -432,12 +432,12 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Sending
             await dispatcher.Dispatch(new TransportOperations(operations.ToArray()), azureServiceBusTransaction.TransportTransaction);
 
             Assert.That(transactionalSender.BatchSentMessages, Has.Count.EqualTo(3));
-            var someDestinationTransactionalBatchContent1 = transactionalSender[transactionalSender.BatchSentMessages.ElementAt(0)];
-            var someDestinationTransactionalBatchContent2 = transactionalSender[transactionalSender.BatchSentMessages.ElementAt(1)];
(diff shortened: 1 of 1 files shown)


src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -128,6 +128,7 @@ async Task DispatchBatchForDestination(string destination, ServiceBusClient? cli
             var messageCount = messagesToSend.Count;
             int batchCount = 0;
             var sender = messageSenderRegistry.GetMessageSender(destination, client);
+            int maxItemsPerBatch = transaction == null ? int.MaxValue : 100;
```
danielmarbach, line 131: There is a magic limit of 4500 messages for non-transactional sends. That magic limit is depending on the tier and should be incorporated at some point into the `TryAdd` method. That's why I chose `int.Max` for now. 

https://github.com/Azure/azure-sdk-for-net/issues/21451

Happy to also set it to 4500 in the non-tx case if you prefer that.

danielmarbach: Realistically, this limit of 4500 probably cannot be reached with the amount of data that is added to a single message. Even if the body is empty, the headers are still big enough to make sure we never reach, 4500 msgs. I did a quick test with sending 4500 empty messages and it ended up batching roughly 250 into one single batch due to the size constraints.

lailabougria (approved):

PR #670, opened 2022-11-04
Introduce the topic topology to enable subscription to multiple topics

danielmarbach:
This PR introduces the possibility to circumvent the limitations of the current forwarding topology by separating the publish topic from the subscribe topic.

Currently the limitations of the topology are described in our [topology documentation](https://docs.particular.net/transports/azure-service-bus/topology#quotas-and-limitations). At first sight, it looks like we can support quite a few endpoints. In reality, this number is much smaller because the more event types you get, the more likely it is that you are running into the limit of 2000 rules since every event type SQL filter is represented as its own rule.

Relevant [quotas](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-quotas) for the discussion

image

As an example, in an event-driven system you may have dozens of event types for various business components to interact with each other. Just for the example, let's assume we are dealing with a system of 100 different event types. To simplify the math, let's assume every endpoint available will be on average subscribe to 25% of the events. That means every endpoint has to manage at least 25 rules, which brings us down to a scaling limit of 80 endpoints which is definitely far away from the 2000 endpoints as the documentation suggests. The more events you have and the more and endpoint is interested in subscribing to them, you get less and fewer endpoints you can start adding to your system. 

Currently, when you start approaching those limits, it is only possible to go the extreme route of managing the whole topology yourself. This would mean you have to:

1. Disable auto subscription on all the endpoints
2. Segregate all endpoints into different topics
3. Have bridges interacting between the different segments of the system

This manual management is error-prone and cumbersome. This PR attempts to solve that.

## Existing systems

By default the transport uses `bundle-1` for the publish topic and the subscribe topic. This makes everything compatible with existing endpoints. Once you start approaching the limits of the transport it is possible add new subscribers on a new subscription topic but still publish to `bundle-1`. The endpoint will then automatically add a forwarding rule.

Topology

Existing Subscriber 1 and Subscriber 2 can be left as is. Subscriber 3 has to define the hierarchy.

```csharp
// Subscriber3
var transport = b.ConfigureTransport<AzureServiceBusTransport>();
transport.Topology = TopicTopology.Hierarchy("bundle-1", "bundle-2");
```

Should the existing system use something else than `bundle-1` like the following code illustrates

```csharp
// Subscriber1
var transport = b.ConfigureTransport<AzureServiceBusTransport>();
transport.TopicName = "my-custom-bundle";

// Subscriber2
var transport = b.ConfigureTransport<AzureServiceBusTransport>();
transport.TopicName = "my-custom-bundle";
```

Subscriber 3 would have to define the hierarchy like

```csharp
// Subscriber3
var transport = b.ConfigureTransport<AzureServiceBusTransport>();
transport.Topology = TopicTopology.Hierarchy("my-custom-bundle", "my-custom-sub-bundle");
```

## New systems

Can choose freely how the want to segregate things into different "bundle topics" or start with the defaults and then later switch as described above.

## Caveat

- The maximum number of forwarding hops is 4. The "sub" topics increase the hops by one but it still falls into the limit. 
- In theory things could be nested further as long as it fits into the maximum number of 4 hops.
- For standard namespaces this will increase slightly the costs because forwarding operations will also be charged. Premium namespace users are billed per messaging unit (which is CPU and memory) and are not affected.

## Missing part

- [x] Making the CLI tool support this


abparticular (approved):

danielmarbach:
@abparticular @lailabougria can you please give this a final review?

src/CommandLine/Endpoint.cs
```diff
@@ -8,7 +8,7 @@
 
     static class Endpoint
     {
-        public static async Task Create(ServiceBusAdministrationClient client, CommandArgument name, CommandOption topicName, CommandOption subscriptionName, CommandOption<int> size, CommandOption partitioning)
+        public static async Task Create(ServiceBusAdministrationClient client, CommandArgument name, CommandOption topicName, CommandOption topicToPublishTo, CommandOption topicToSubscribeOn, CommandOption subscriptionName, CommandOption<int> size, CommandOption partitioning)
```
abparticular, line 11: Given that TopicName is being marked as Obsolete in AzureServiceBus transport, is it worth having two different overloads here - one with just topic name (marked as obsolete) and one with publish/subscribe topics?  The thought being that once TopicName is removed from the transport, it will be easier to remove the corresponding functionality here.

abparticular (approved):

src/CommandLine/Endpoint.cs
danielmarbach: @lailabougria Andreas and I had a conversation about this. I told him the idea of this change is to make sure the existing `topicName` option can still be used and represents the `single` topic topology while the other two parameters represent the hierarchy options. Validation will take care those parameters can only be used in valid combinations. Once validation is over I could theoretically map the single topic name parameter to the publish/subscribe topics but then I would create an artificial state of these parameters that validation previously made sure it should not be possible which also made me a bit iffy. I'm not opposed changing this code if we can come up with a better solution but I'm also not too overly worried if we leave it like that since those API parameter closely follow the assumptions of the command line usage anyway.

src/CommandLine/Program.cs
```diff
@@ -176,6 +183,29 @@ static int Main(string[] args)
             }
         }
 
+        static ValidationResult ValidateTopicInformationIsCorrect(CommandOption topicName, CommandOption topicToPublishTo, CommandOption topicToSubscribeOn)
+        {
+            if (topicName.HasValue() && topicToPublishTo.HasValue())
+            {
+                return new ValidationResult(
+                    "The topic name and the topic to publish to option cannot be used together. Choose either a single topic name by specifying the topic name or a hierarchy by specifying the topic to publish to and the topic to subscribe on.");
```
lailabougria, line 191:
```suggestion
                    "The --topic option and the --topic-to-publish-to option cannot be combined. Choose either a single topic name by specifying the --topic option or a hierarchy by specifying both the --topic-to-publish-to option, and --topic-to-subscribe-on option.");
```

src/CommandLine/Program.cs
```diff
@@ -176,6 +183,29 @@ static int Main(string[] args)
...
+            }
+
+            if (topicName.HasValue() && topicToSubscribeOn.HasValue())
+            {
+                return new ValidationResult(
+                    "The topic name and the topic to subscribe on option cannot be used together. Choose either a single topic name by specifying the topic name or a hierarchy by specifying the topic to publish to and the topic to subscribe on.");
```
lailabougria, line 197:
```suggestion
                return new ValidationResult(
                    "The --topic option and the--topic-to-subscribe-on option cannot be combined. Choose either a single topic name by specifying the --topic option or a hierarchy by specifying both the --topic-to-publish-to option, and --topic-to-subscribe-on option.");
```

src/CommandLine/Program.cs
```diff
@@ -176,6 +183,29 @@ static int Main(string[] args)
...
+            }
+
+            if (topicToPublishTo.HasValue() && topicToSubscribeOn.HasValue() && string.Equals(topicToPublishTo.Value(), topicToSubscribeOn.Value(), StringComparison.OrdinalIgnoreCase))
+            {
+                return new ValidationResult(
+                    "In order to represent a topic hierarchy the topic to publish to and the topic to subscribe on need to be different.");
```
lailabougria, line 203:
```suggestion
                    "A valid topic hierarchy requires the topic-to-publish-to option and the topic-to-subscribe-on option to be different.");
```

src/CommandLineTests/CommandLineTests.cs
```diff
@@ -42,6 +64,27 @@ public async Task Create_endpoint_validates_namespace_and_connection_string_cann
...
 
+        [Test]
+        public async Task Create_endpoint_validates_topic_name_and_hierarchy_cannot_be_used_together()
+        {
+            var (_, publishError, publishExitCode) = await Execute($"endpoint create {EndpointName} --topic {TopicName} --topic-to-publish-to {HierarchyTopicName}");
+            var (_, subscribeError, subscribeExitCode) = await Execute($"endpoint create {EndpointName} --topic {TopicName} --topic-to-subscribe-on {HierarchyTopicName}");
+
+            Assert.AreEqual(1, publishExitCode);
+            Assert.AreEqual(1, subscribeExitCode);
+            StringAssert.Contains("The topic name and the topic to publish to option cannot be used together.", publishError);
```
lailabougria, line 75:
```suggestion
            StringAssert.Contains("The --topic option and the --topic-to-publish-to option cannot be combined. Choose either a single topic name by specifying the --topic option or a hierarchy by specifying both the --topic-to-publish-to option, and --topic-to-subscribe-on option.", publishError);
```

src/CommandLineTests/CommandLineTests.cs
```diff
@@ -42,6 +64,27 @@ public async Task Create_endpoint_validates_namespace_and_connection_string_cann
...
+            StringAssert.Contains("The topic name and the topic to subscribe on option cannot be used together.", subscribeError);
```
lailabougria, line 76:
```suggestion
            StringAssert.Contains("The --topic option and the--topic-to-subscribe-on option cannot be combined. Choose either a single topic name by specifying the --topic option or a hierarchy by specifying both the --topic-to-publish-to option, and --topic-to-subscribe-on option.", subscribeError);
```

src/CommandLineTests/CommandLineTests.cs
```diff
@@ -42,6 +64,27 @@ public async Task Create_endpoint_validates_namespace_and_connection_string_cann
...
+        }
+
+        [Test]
+        public async Task Create_endpoint_validates_hierarchy_is_correct()
+        {
+            var (_, error, exitCode) = await Execute($"endpoint create {EndpointName} --topic-to-subscribe-on {HierarchyTopicName} --topic-to-publish-to {HierarchyTopicName}");
+
+            Assert.AreEqual(1, exitCode);
+            StringAssert.Contains("In order to represent a topic hierarchy the topic to publish to and the topic to subscribe on need to be different.", error);
```
lailabougria, line 85:
```suggestion
            StringAssert.Contains("A valid topic hierarchy requires the topic-to-publish-to option and the topic-to-subscribe-on option to be different.", error);
```

lailabougria (approved):
Nitpicks, but I thought it would be useful that the error messages point to the option names for readability on the user end.

PR #695, opened 2022-11-21
Initialize SubscriptionManager outside MessagePump

timbussmann:
Small refactoring to remove some unnecessary dependencies from the `MessagePump` constructor

### DIFF
--- a/src/Tests/Receiving/MessagePumpTests.cs
+++ b/src/Tests/Receiving/MessagePumpTests.cs
@@ -16,7 +16,7 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Receiving
             var fakeClient = new FakeServiceBusClient();
             var fakeReceiver = new FakeReceiver();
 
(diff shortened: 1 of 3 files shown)


danielmarbach (approved):

danielmarbach:
I wonder if we should have a No-Op Subscription manager instead of Null but that is really just nitpicking.

timbussmann:
> I wonder if we should have a No-Op Subscription manager instead of Null but that is really just nitpicking.

Code-wise, it would make sense. This is a bit of a rough edge of the transport seam since we couldn't use nullable reference types yet to mark this to be potentially null. The idea was to indicate to the receiver/pump whether that instance supports pub/sub in the first place. The `Subscriptions` being null kind of communicates that (in a shitty way), while a null instance would require a bit more internal knowledge for checks as you couldn't rely on simple null checks anymore but are required to check for a specific (core provided) null instance or type. Might be worth the trade-off in case we prefer avoiding the risk of NREs for consumers of the API over preventing invalid usage of the API (subscribing when it's not possible).

PR #750, opened 2023-01-12
Pause longer when circuit breaker has been triggered

SzymonPobiega:
This is meant to address two issues
 - https://github.com/Particular/ServiceControl/issues/2907
 - https://github.com/Particular/ServiceControl/issues/3318
 
by checking the state of the circuit breaker when `Failure` is invoked. If the breaker has already been triggered, the method pauses for 10 seconds (instead of 1 second) to prevent flooding the logs. The issue affects both ServiceControl and regular endpoints.

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -94,7 +94,20 @@
 
             messageProcessingCancellationTokenSource = new CancellationTokenSource();
 
-            circuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker($"'{receiveSettings.ReceiveAddress}'", transportSettings.TimeToWaitBeforeTriggeringCircuitBreaker, ex => criticalErrorAction("Failed to receive message from Azure Service Bus.", ex, messageProcessingCancellationTokenSource.Token));
+            circuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker($"'{receiveSettings.ReceiveAddress}'",
+                transportSettings.TimeToWaitBeforeTriggeringCircuitBreaker, ex =>
+                {
+                    criticalErrorAction("Failed to receive message from Azure Service Bus.", ex,
+                        messageProcessingCancellationTokenSource.Token);
+                }, () =>
+                {
+                    //We don't have to update the prefetch count since we are failing to receive anyway
+                    processor.UpdateConcurrency(1);
+                },
+                () =>
+                {
+                    processor.UpdateConcurrency(limitations.MaxConcurrency);
+                });
 
             await processor.StartProcessingAsync(cancellationToken)
                 .ConfigureAwait(false);
--- a/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
+++ b/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
@@ -25,6 +25,7 @@
                 return;
             }
 
+            triggered = false;
             timer.Change(Timeout.Infinite, Timeout.Infinite);
             Logger.InfoFormat("The circuit breaker for {0} is now disarmed", name);
         }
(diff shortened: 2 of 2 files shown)


jpalac (approved):

WojcikMike (approved):

src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -40,7 +41,10 @@ public Task Failure(Exception exception, CancellationToken cancellationToken = d
                 Logger.WarnFormat("The circuit breaker for {0} is now in the armed state", name);
             }
 
-            return Task.Delay(TimeSpan.FromSeconds(1), cancellationToken);
+            //If the circuit breaker has been triggered, wait for 10 seconds before proceeding to prevent flooding the logs and hammering the ServiceBus
+            var delay = triggered ? TimeSpan.FromSeconds(10) : TimeSpan.FromSeconds(1);
```
tmasternak, line 45: Why do we want to increase the delay only after the circuit has been triggered? Why not always go with `10` seconds delay?

tmasternak (changes requested):

src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
SzymonPobiega: Good question. Maybe that would be a good enough solution. What do you think @WojcikMike & @jpalac ?

WojcikMike: I don't mind that. I was checking other transports, but there are different solutions. So increasing the time to 10 secs is fine

jpalac: Do the other solutions not incorporate a timer? Are they worth looking at to use similar logic or are the transports too different that it's not worth it?

WojcikMike: Rabbit is using this: https://github.com/Particular/NServiceBus.RabbitMQ/blob/master/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePumpConnectionFailedCircuitBreaker.cs

And the value is configurable: 
https://docs.particular.net/transports/rabbitmq/connection-settings#controlling-behavior-when-the-broker-connection-is-lost-timetowaitbeforetriggering
Default to 2 mins

The same for SqlServer: 
https://docs.particular.net/transports/sql/connection-settings?version=sqltransport_6#circuit-breaker-wait-time

tmasternak (approved):

SzymonPobiega:
I updated the PR so now when the cb is triggered the concurrency is changed to 1 to limit the log flooding even more.

src/Transport/Receiving/MessagePump.cs
```diff
@@ -94,7 +94,20 @@ public async Task StartReceive(CancellationToken cancellationToken = default)
...
-            circuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker($"'{receiveSettings.ReceiveAddress}'", transportSettings.TimeToWaitBeforeTriggeringCircuitBreaker, ex => criticalErrorAction("Failed to receive message from Azure Service Bus.", ex, messageProcessingCancellationTokenSource.Token));
+            circuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker($"'{receiveSettings.ReceiveAddress}'",
+                transportSettings.TimeToWaitBeforeTriggeringCircuitBreaker, ex =>
+                {
+                    criticalErrorAction("Failed to receive message from Azure Service Bus.", ex,
+                        messageProcessingCancellationTokenSource.Token);
+                }, () =>
+                {
+                    //We don't have to update the prefetch count since we are failing to receive anyway
+                    processor.UpdateConcurrency(1);
```
tmasternak, line 105: I was wondering if there is some overlap with changes from [#761](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/761/files#diff-bf785c1b37e39481e89856ea4195ca5c7e21f862b8eb2ffc43b63208a3c582e5R207) but I think we are good.

tmasternak (approved):

WojcikMike (approved):

PR #761, opened 2023-01-19
Switch the transport to use UpdatePrefetchCount on the processor when rate limiting occurs

danielmarbach:
This leverages the feature [I contributed to the Azure.Messaging.ServiceBus library](https://github.com/Azure/azure-sdk-for-net/pull/33162/) that was released as part of 7.12.0. With that approach, it is no longer necessary to stop and start the transport. We can now adjust the processor's concurrency and prefetch count. By setting the prefetch count the underlying AMQP link credit is updated automatically if it has changed.

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -65,12 +65,7 @@
 
         public async Task StartReceive(CancellationToken cancellationToken = default)
         {
-            var prefetchCount = limitations.MaxConcurrency * transportSettings.PrefetchMultiplier;
-
-            if (transportSettings.PrefetchCount.HasValue)
-            {
-                prefetchCount = transportSettings.PrefetchCount.Value;
-            }
+            int prefetchCount = CalculatePrefetchCount();
 
             var receiveOptions = new ServiceBusProcessorOptions
             {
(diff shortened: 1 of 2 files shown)


src/Transport/Receiving/MessagePump.cs
```diff
@@ -197,26 +204,16 @@ await circuitBreaker.Failure(processErrorEventArgs.Exception, processErrorEventA
...
-                // For all other cases the users is using either the default multiplier or a defined multiplier
-                // that sets the prefetch count in accordance of the maximum concurrency. In those scenarios we cannot
-                // use UpdateConcurrency because that would not adjust the prefetch count to the new desired values
-                // therefore we are stopping and restarting (which also creates a new underlying AMQP link that will have
-                // the new prefetch count settings.
-                await StopReceive(cancellationToken).ConfigureAwait(false);
-                await StartReceive(cancellationToken).ConfigureAwait(false);
-            }
+
+            processor.UpdateConcurrency(limitations.MaxConcurrency);
```
tmasternak, line 211: Do we need that considering that we set the concurrency in [line 82](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/761/files#diff-bf785c1b37e39481e89856ea4195ca5c7e21f862b8eb2ffc43b63208a3c582e5R77)?

SzymonPobiega (approved):

src/Transport/Receiving/MessagePump.cs
danielmarbach: The code that sets the concurrency is the code that takes the value from the endpoint configuration. Let's say a user has `LimitProcessingConcurrencyTo(15)` then it sets the maximum concurrency to `15` in the code you mentioned. When the core rate-limiting kicks in it sets the concurrency to `1` until the failures are gone and then sets the concurrency back to `15` in this example. The code here sets the concurrency of the processor from one value to another as instructed by the core.

Or maybe I'm misunderstanding you.

tmasternak (approved):

PR #768, opened 2023-02-23
Load exception on debug log entry

WilliamBZA:
Logs the values of all the relevant variables if a null-ref exception occurs on writing a debug statement.


DavidBoike:
LGTM, only thing I could think to add would be…

```cs
static readonly DateTime start = DateTime.UtcNow;
```

…and then the log message could include `Uptime: {DateTime.UtcNow - start}` which might help prove or disprove the happening on startup when the log switch happens theory.

I don't really need to review though, as we don't need to merge. Can get the PR build out of GitHub Actions as long as it's in the 24h before that gets thrown away.

WilliamBZA:
I added the start time log

PR #779, opened 2023-03-09
Manage subscriptions without manage rights

### LINKED ISSUE #645: Switch where possible the SubscriptionManager to use the RuleManager
The SubscriptionManager uses internally an admin client to create and update rules. The Azure Service Bus SDK has a new type that is called `ServiceBusRuleManager` that can be acquired with

```csharp
            await using var client = new ServiceBusClient(connectionString);
            var ruleManager = client.CreateRuleManager(topicName, subscriptionName);
```

the rule manager allows creating and deleting subscriptions without requiring to manage claims. This would be a huge win because we can allow subscribing and unsubscribing at runtime on an already created subscription without requiring "admin rights" which allows running larger portions of the transport following the least privilege principle. 

Presently, the subscription manager has [slightly twisted logic](#567 ) that updates rules and updating rules is [currently not supported on the rule manager](https://github.com/Azure/azure-sdk-for-net/issues/31625).

danielmarbach:
Closes #645 

This PR leverages the rule manager. The rule manager [doesn't support updating rules](https://github.com/Azure/azure-sdk-for-net/issues/31625), so we have to first delete the rules and then create them. It is also not possible to [enumerate rules on subscriptions with forwarding enabled](https://github.com/Azure/azure-sdk-for-net/pull/34816).

### DIFF
--- a/src/Transport/Administration/SubscriptionManager.cs
+++ b/src/Transport/Administration/SubscriptionManager.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Transport.AzureServiceBus
 {
     using System;
+    using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
     using Azure.Messaging.ServiceBus;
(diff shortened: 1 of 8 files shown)


src/Transport/Administration/SubscriptionManager.cs
```diff
@@ -14,83 +15,99 @@ class SubscriptionManager : ISubscriptionManager
...
+            // to preserve the update or create behavior
 
             try
             {
-                var existingRule = await client.GetRuleAsync(transportSettings.Topology.TopicToSubscribeOn, subscriptionName, rule.Name, cancellationToken).ConfigureAwait(false);
-
-                if (existingRule.Value.Filter.ToString() != rule.Filter.ToString())
-                {
-                    existingRule.Value.Action = rule.Action;
+                await ruleManager.DeleteRuleAsync(ruleName, cancellationToken)
```
andreasohlund, line 73: I get why we have to delete first based on the comment above but this feels like a potential for message loss since:

1. Now there will be a short window of time during each enpoint restart where the subscription is not in place
2. If something fails when we try to re-create the rule on 83 that failure would mean that we could potentially leave the subscription missing until the thing causing the failure is resolved

Am I reading this the wrong way?

danielmarbach: Yeah I wanted to discuss this with you in more details

src/Transport/Administration/SubscriptionManager.cs
```diff
@@ -14,83 +15,88 @@ class SubscriptionManager : ISubscriptionManager
...
+            {
+                return;
+            }
 
-            foreach (var eventType in eventTypes)
+            var ruleManager = client.CreateRuleManager(transportSettings.Topology.TopicToSubscribeOn, subscriptionName);
+            await using (ruleManager.ConfigureAwait(false))
             {
-                await SubscribeEvent(administrativeClient, eventType.MessageType, cancellationToken).ConfigureAwait(false);
+                if (eventTypes.Length == 1)
```
andreasohlund, line 44: So this is to avoid a performance hit when using Task.WhenAll? (Is it worth it given that we only execute this once at startup)

andreasohlund (approved):

src/Transport/Administration/SubscriptionManager.cs
danielmarbach: Subscribe is called during runtime as well with an array with one entry.

PR #835, opened 2023-06-26
Remove sender pooling

mikeminutillo:
The sender pool is not necessary with the supported versions of the client SDK and can lead to a higher than expected number of senders.

The message sender registry was initially implemented in #644


danielmarbach (approved):

danielmarbach:
Probably squash my silly commits :D

lailabougria (approved):

PR #839, opened 2023-06-28
Detect oversized batch and throw

mikeminutillo:
If you try to send more than 100 messages in a single transaction Azure Service Bus will throw an exception https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-quotas

When that happens the incoming message is not handled properly https://github.com/Azure/azure-sdk-for-net/issues/37265

This change detects that you are trying to do this and does not attempt any dispatches.


mikeminutillo:
Raised #840 as a simpler option

mikeminutillo:
Closed in favor of #840

PR #840, opened 2023-06-29
Simpler detect oversized batch

mikeminutillo:
- Alternative to #839 

Checks for attempts to send more than 100 messages in a single transaction, which is not allowed https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-quotas and throws an exception.

Compared to #839 this PR makes minimal changes to the dispatcher. It iterates through the outgoing transactions only if the total count is high enough to potentially go over th limit, and if we are participating in a transaction.

### DIFF
--- a/src/Transport/Sending/MessageDispatcher.cs
+++ b/src/Transport/Sending/MessageDispatcher.cs
@@ -71,10 +71,14 @@
             return tasks.Count == 1 ? tasks[0] : Task.WhenAll(tasks);
         }
 
-        static void AssertBelowMaxMessageThresholdForTransaction(List<UnicastTransportOperation> unicastTransportOperations, List<MulticastTransportOperation> multicastTransportOperations, CommittableTransaction committableTransaction)
+        /// <summary>
+        /// Throws an exception if attempting to send more than 100 messages in a transaction.
+        /// This prevents the transport from experiencing this issue https://github.com/Azure/azure-sdk-for-net/issues/37265
+        /// </summary>
+        static void AssertBelowMaxMessageThresholdForTransaction(List<UnicastTransportOperation> unicastTransportOperations, List<MulticastTransportOperation> multicastTransportOperations, Transaction transaction)
         {
             var totalNumberOfOperations = unicastTransportOperations.Count + multicastTransportOperations.Count;
-            if (committableTransaction == null || totalNumberOfOperations <= MaxMessageThresholdForTransaction)
+            if (transaction == null || totalNumberOfOperations <= MaxMessageThresholdForTransaction)
             {
                 return;
             }
--- a/src/Transport/Sending/MessageDispatcher.cs
+++ b/src/Transport/Sending/MessageDispatcher.cs
@@ -1,5 +1,6 @@
 ﻿namespace NServiceBus.Transport.AzureServiceBus
 {
+    using System;
     using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


danielmarbach (approved):

src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -66,6 +71,36 @@ public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction
             return tasks.Count == 1 ? tasks[0] : Task.WhenAll(tasks);
         }
 
+        static void AssertBelowMaxMessageThresholdForTransaction(List<UnicastTransportOperation> unicastTransportOperations, List<MulticastTransportOperation> multicastTransportOperations, CommittableTransaction committableTransaction)
```
danielmarbach, line 74: Nit; Transaction could be of the base type here

src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -66,6 +71,36 @@ public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction
...
+        {
+            var totalNumberOfOperations = unicastTransportOperations.Count + multicastTransportOperations.Count;
+            if (committableTransaction == null || totalNumberOfOperations <= MaxMessageThresholdForTransaction)
+            {
+                return;
+            }
+
+            var numberOfTransactionalOperations = 0;
+            foreach (var transportOperation in unicastTransportOperations)
```
danielmarbach, line 83: Should we be worried about the double enumeration we now have? I think it could be ok because we are only doing it in a very specific niche scenario to add more safety

danielmarbach: I wonder why we can't directly count and assert in existing loops. We have all the information available like transaction and operation count, the dispatch type etc

mikeminutillo: If we do that, we need to remove the dispatch from the loop. I suppose we could still dispatch the first 100 and then throw when / if we get to the limit.

danielmarbach: I think it is better not do any IO you are right because that is more expensive that double iteration. Let's leave it

mikeminutillo: Raised https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/841 to do that

danielmarbach: Did you see https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/840#discussion_r1246111434 😁?

mikeminutillo: Not until after I had done the work :)

danielmarbach, line 83:
```suggestion
             // we are iterating through the operations twice, the tradeoff here is that we rather enumerate twice
             // than incurring IO costs in scenarios we know it will fail
            var numberOfTransactionalOperations = 0;
            foreach (var transportOperation in unicastTransportOperations)
```

lailabougria: @danielmarbach Since this is comment refers to the other alternatives, we prefer to document this in an ADR.

PR #841, opened 2023-06-29
Detect oversized batch 3

mikeminutillo:
- Alternative to #840
- Alternative to #839

This version counts transactional operations as we go and throws once we exceed the limit. This means that we will try and dispatch the first 100 messages and throw when we try to dispatch the 101st


danielmarbach:
I added a suggestion to https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/840 which is my preferred option because it avoids extra IO

mikeminutillo:
Closed in favor of #840

PR #884, opened 2023-09-22
[DO NOT MERGE] Spike correlation filter support

mikeminutillo:
Enables overriding the default subscription rule filter and runs the tests against correlation filters.

This is probably not how we would do this, but it should hopefully demonstrate that everything still works.


SeanFeldman:
Correlation filters won’t work with [polymorphic dispatching](https://docs.particular.net/nservicebus/messaging/messages-as-interfaces).

danielmarbach:
@SeanFeldman we are aware of that. Thanks for chiming in!

mikeminutillo:
We know. It's why we're being cautious about adding it. There's a bunch of caveats so it's only safe if you know exactly what it's doing.

SeanFeldman:
> This is probably not how we would do this, but it should hopefully demonstrate that everything still works.

That statement in the description is what triggered my comment. Otherwise, I'm sure you know what you know 🙂

PR #980, opened 2024-05-10
Provide an additional point for users to troubleshoot max message size exception

SeanFeldman:
When the maximum message size is changed but endpoints are not restarted, ASB client holds on to the old maximum message size, causing the transport to fail on the messages that otherwise would be OK to send using safe batching API.

SDK issue: https://github.com/Azure/azure-sdk-for-net/issues/43983


andreasohlund:
Thanks @SeanFeldman I've pulled this in via https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/981

PR #993, opened 2024-05-27
DefaultCredential acceptance tests

danielmarbach:
Should be squashed

### DIFF
--- a/src/AcceptanceTests/When_using_token_credential_with_fully_qualified_namespace.cs
+++ b/src/AcceptanceTests/When_using_token_credential_with_fully_qualified_namespace.cs
@@ -1,0 +1,87 @@
+namespace NServiceBus.Transport.AzureServiceBus.AcceptanceTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Azure.Identity;
+    using Azure.Messaging.ServiceBus;
+    using NServiceBus.AcceptanceTests;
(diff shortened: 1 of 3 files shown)


PhilBastian (approved):

src/AcceptanceTests/When_using_token_credential_with_fully_qualified_namespace.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        [Test]
+        public async Task Should_work()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Publisher>(b =>
+                {
+                    b.CustomConfig(c =>
+                    {
+                        var transport = c.ConfigureTransport<AzureServiceBusTransport>();
```
andreasohlund, line 32: Does our test framework not allow the new API to be used? (`c.ConfigureTransport(new AzureServiceBusTransport(fqns, credential))`

andreasohlund (approved):

src/AcceptanceTests/When_using_token_credential_with_fully_qualified_namespace.cs
danielmarbach: Yeah I will fix this

danielmarbach: Actually, I think we have to use that one; otherwise we are essentially setting a new transport definition and would not get the default values from the template if I recall correctly

PR #996, opened 2024-06-07
Spike correlation filter support

ramonsmits:
Replaces:

- https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/884

### DIFF
--- a/src/TransportTests/ConfigureAzureServiceBusTransportInfrastructure.cs
+++ b/src/TransportTests/ConfigureAzureServiceBusTransportInfrastructure.cs
@@ -3,6 +3,7 @@ using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Azure.Messaging.ServiceBus.Administration;
 using NServiceBus;
 using NServiceBus.Transport;
 using NServiceBus.TransportTests;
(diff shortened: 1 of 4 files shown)


src/TransportTests/ConfigureAzureServiceBusTransportInfrastructure.cs
```diff
@@ -39,7 +40,8 @@ string HexStringFromBytes(byte[] bytes)
                         return sb.ToString();
                     }
                 }
-            }
+            },
+            SubscriptionRuleFilterConvention = eventType => new CorrelationRuleFilter(eventType.FullName)
```
SeanFeldman, line 44: Is the `Headers.EnclosedMessageTypes` value going to be propagated to the message's `CorrelationId` property?

ramonsmits: @SeanFeldman This is only a rebase of:

- https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/884

From what I read correlation filter work on application specific message properties too?

Why should it be assigned to the `CorrelationId` property?

sean-feldman-sh: > From what I read correlation filter work on application specific message properties too?

Correct.

> Why should it be assigned to the CorrelationId property?

You're setting a value (message type) through the constructor. That value by default will be compared against `CorrelationId` system property.

PR #1017, opened 2024-08-05
upgrade to Azure.Messaging.ServiceBus version that limits batch TryAdd when size limit is reached

### LINKED ISSUE #994: An endpoint is "force detached" by the broker when sending messages
#### Description

When sending/publishing messages (either using transactional session or inside a message handler), the default behaviour is to batch them and send them after the method calling the send/publish method has exited. (This behavior can be changed to dispatch immediately, but then transactionality is lost.)

#### Expected behavior

The messages are sent.

#### Actual behavior

The messages are not sent, and an exception is logged.

#### Versions

3.0.0 onwards. Introduced in https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/644

### Steps to reproduce

In a message handler, send some messages, with the total size of all messages greater than the max allowed message size of the queue.

### Relevant log output

```shell
> System.InvalidOperationException: The link 'G6:14740510:amqps://**namespace removed**.servicebus.windows.net/-dd910ad4;0:7:8' is force detached by the broker because publisher(link3082) received a batch message with no data in it. Detach origin: Publisher.
For troubleshooting information, see https://aka.ms/azsdk/net/servicebus/exceptions/troubleshoot.
```

#### Workarounds

None identified so far.

Note that switching to immediate dispatch (actually sending the message when the send/publish method is called) is not a workaround, because that is very different behaviour which loses any guarantees around message sending/publishing transactionality.

#### Possible solutions

- Measure the size of the outgoing batch before sending it, and if it's too large, don't batch.
- Catch the exception, and retry in a non-batched manner. This probably requires some better behaviour from the Azure SDK client, see https://github.com/Azure/azure-sdk-for-net/issues/44261
- Do not fix the bug, but instead provide a configuration switch to switch off batching. This may not be sufficient to fix the bug in ServiceControl (see below).

#### Additional information

This also manifests in ServiceControl when it attempts to forward audit messages. See https://github.com/Particular/ServiceControl/issues/4185.

Our resolution to this bug may well depend on changes to the Azure SDK client, see https://github.com/Azure/azure-sdk-for-net/issues/44261

PhilBastian:
fixes #994

### DIFF
--- a/src/AcceptanceTests/NServiceBus.Transport.AzureServiceBus.AcceptanceTests.csproj
+++ b/src/AcceptanceTests/NServiceBus.Transport.AzureServiceBus.AcceptanceTests.csproj
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.12.0" />
-    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.17.5" />
+    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.18.1" />
     <PackageReference Include="GitHubActionsTestLogger" Version="2.3.3" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.10.0" />
     <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="9.1.0" GeneratePathProperty="true" />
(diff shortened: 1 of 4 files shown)


danielmarbach:
The namespaces get created on the fly during CI/CD runs. A premium namespace is billed per hour and costs roughly 2 USD per hour (a test run runs for a fraction of an hour). The question is though whether we really want to test the SDK?

> It is awesome when all u need to do is update a package

I guess my [asynchronous lobbying](https://github.com/Azure/azure-sdk-for-net/issues/44261) has helped ;)

danielmarbach (approved):

johnsimons:
Yeah agree @danielmarbach 😄

PR #1018, opened 2024-08-06
Remove the 4500 batch limit since that is now enforced by the update SDK reference

danielmarbach:


danielmarbach:
Looking at this a bit closer I think to not introduce breaking changes further work is required. As of today, on the released versions the transport does split more than 4500 messages into multiple batches. If we remove the explicit check for the 4500 combined with the newest SDK updates, `TryAdd` will return `false` which then throws a ServiceBusException. So essentially by relying on the SDK only we would break splitting up into multiple batches.

danielmarbach:
Should probably be replaced with https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1019

PR #1021, opened 2024-08-06
Cancel the pipeline execution when the lock is lost

danielmarbach:
This PR leverages the feature that I contributed to the client library. The client library has the possibility to subscribe to message lock lost events. [When the message lock is lost, the message becomes available for consumption again
](https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-transfers-locks-settlement#peeklock)

> When a message is locked, other clients receiving from the same queue or subscription can take on locks and retrieve the next available messages not under active lock. When the lock on a message is explicitly released or when the lock expires, the message is placed at or near the front of the retrieval order for redelivery.

and attempts to settle the message will fail. See

> The Complete, DeadLetter, or RenewLock operations might fail due to network issues, if the held lock has expired, or there are other service-side conditions that prevent settlement. In one of the latter cases, the service sends a negative acknowledgment that surfaces as an exception in the API clients. If the reason is a broken network connection, the lock is dropped since Service Bus doesn't support recovery of existing AMQP links on a different connection.
> 
> If Complete fails, which occurs typically at the very end of message handling and in some cases after minutes of processing work, the receiving application can decide whether to preserve the state of the work and ignore the same message when it's delivered a second time, or whether to toss out the work result and retries as the message is redelivered.

the same applies for [Abandon](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusreceiver.abandonmessageasync?view=azure-dotnet)

> The lock for the message has expired or the message has already been completed. This does not apply for session-enabled entities. The [Reason](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusexception.reason?view=azure-dotnet#azure-messaging-servicebus-servicebusexception-reason) will be set to [MessageLockLost](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusfailurereason?view=azure-dotnet#azure-messaging-servicebus-servicebusfailurereason-messagelocklost) in this case.

While this mechanism only applies for the PeekLock Mode the client will simply not raise those events in the Receive-And-Delete mode.

By triggering the cancellation token that is eventually passed into Core when the message lock is lost this gives the clients that are forwarding the token the opportunity to abort their processing logic earlier since any attempts to eventually complete the message would throw a lock lost and the message would be retried anyway.

### DIFF
--- a/src/Tests/Receiving/MessagePumpTests.cs
+++ b/src/Tests/Receiving/MessagePumpTests.cs
@@ -70,11 +70,13 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Receiving
             var pump = new MessagePump(fakeClient, new AzureServiceBusTransport(), "receiveAddress",
                 new ReceiveSettings("TestReceiver", new QueueAddress("receiveAddress"), false, false, "error"), (s, exception, arg3) => { }, null);
 
-            bool pumpWasCalled = false;
+            using var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+            var pumpExecutingTaskCompletionSource = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+            await using var _ = cancellationTokenSource.Token.Register(() => pumpExecutingTaskCompletionSource.TrySetCanceled());
 
             await pump.Initialize(new PushRuntimeSettings(1), async (context, token) =>
                 {
-                    pumpWasCalled = true;
+                    pumpExecutingTaskCompletionSource.TrySetResult();
                     await Task.Delay(Timeout.InfiniteTimeSpan, token);
                 },
                 (context, token) => Task.FromResult(ErrorHandleResult.Handled), CancellationToken.None);
@@ -84,13 +86,13 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Receiving
 
             var fakeProcessor = fakeClient.Processors["receiveAddress"];
             var processingTask = fakeProcessor.ProcessMessage(receivedMessage, fakeReceiver);
+            await pumpExecutingTaskCompletionSource.Task;
+
             await fakeProcessor.RaiseMessageLockLost(receivedMessage, new MessageLockLostEventArgs(receivedMessage, new ServiceBusException("Lock Lost", ServiceBusFailureReason.MessageLockLost)));
 
             Assert.ThrowsAsync<TaskCanceledException>(async () => await processingTask);
-
             Assert.That(fakeReceiver.CompletedMessages, Is.Empty);
             Assert.That(fakeReceiver.AbandonedMessages, Is.Empty);
-            Assert.That(pumpWasCalled, Is.True);
         }
 
         [Test]
--- a/src/Tests/Receiving/MessagePumpTests.cs
+++ b/src/Tests/Receiving/MessagePumpTests.cs
@@ -62,6 +62,38 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Receiving
         }
 
         [Test]
+        public async Task Should_cancel_receive_token_when_lock_expires_during_processing()
+        {
+            var fakeClient = new FakeServiceBusClient();
+            var fakeReceiver = new FakeReceiver();
+
+            var pump = new MessagePump(fakeClient, new AzureServiceBusTransport(), "receiveAddress",
+                new ReceiveSettings("TestReceiver", new QueueAddress("receiveAddress"), false, false, "error"), (s, exception, arg3) => { }, null);
+
+            bool pumpWasCalled = false;
+
+            await pump.Initialize(new PushRuntimeSettings(1), async (context, token) =>
+                {
+                    pumpWasCalled = true;
+                    await Task.Delay(Timeout.InfiniteTimeSpan, token);
+                },
+                (context, token) => Task.FromResult(ErrorHandleResult.Handled), CancellationToken.None);
+            await pump.StartReceive();
+
+            var receivedMessage = ServiceBusModelFactory.ServiceBusReceivedMessage(messageId: "SomeId", lockedUntil: DateTimeOffset.UtcNow.AddSeconds(60));
+
+            var fakeProcessor = fakeClient.Processors["receiveAddress"];
+            var processingTask = fakeProcessor.ProcessMessage(receivedMessage, fakeReceiver);
+            await fakeProcessor.RaiseMessageLockLost(receivedMessage, new MessageLockLostEventArgs(receivedMessage, new ServiceBusException("Lock Lost", ServiceBusFailureReason.MessageLockLost)));
+
+            Assert.ThrowsAsync<TaskCanceledException>(async () => await processingTask);
+
+            Assert.That(fakeReceiver.CompletedMessages, Is.Empty);
+            Assert.That(fakeReceiver.AbandonedMessages, Is.Empty);
+            Assert.That(pumpWasCalled, Is.True);
+        }
+
+        [Test]
         public async Task Should_abandon_message_upon_failure_with_retry_required()
         {
             var fakeClient = new FakeServiceBusClient();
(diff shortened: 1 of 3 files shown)


src/Transport/Receiving/MessagePump.cs
```diff
@@ -322,33 +327,54 @@ await processor.CloseAsync(cancellationToken)
...
+                catch (Exception onErrorEx) when (onErrorEx.IsCausedBy(processingTokenSource.Token))
                 {
                     throw;
                 }
                 catch (Exception onErrorEx)
                 {
-                    criticalErrorAction($"Failed to execute recoverability policy for message with native ID: `{message.MessageId}`", onErrorEx, messageProcessingCancellationToken);
+                    criticalErrorAction(
+                        $"Failed to execute recoverability policy for message with native ID: `{message.MessageId}`",
+                        onErrorEx, processingTokenSource.Token);
```
danielmarbach, line 352: This is the place where I wasn't sure what token to add. You could argue that here it might make sense to add the `messageProcessingCancellationToken` so that the critical error action is only cancelled when the endpoint shuts down.

src/Transport/Receiving/MessagePump.cs
```diff
@@ -272,28 +272,30 @@ await processor.CloseAsync(cancellationToken)
...
+                using var azureServiceBusTransaction = CreateTransaction(message.PartitionKey);
+                contextBag.Set(message);
+                contextBag.Set(processMessageEventArgs);
 
-                    var messageContext = new MessageContext(messageId, headers, body, azureServiceBusTransaction.TransportTransaction, ReceiveAddress, contextBag);
+                var messageContext = new MessageContext(messageId, headers, body,
+                    azureServiceBusTransaction.TransportTransaction, ReceiveAddress, contextBag);
 
-                    await onMessage(messageContext, messageProcessingCancellationToken).ConfigureAwait(false);
+                await onMessage(messageContext, processingTokenSource.Token).ConfigureAwait(false);
```
johnsimons, line 288: it may make sense to elevate this into its own `var`
`var messageProcessingCancellationToken = processingTokenSource.Token;`

src/Transport/Receiving/MessagePump.cs
```diff
@@ -322,33 +325,56 @@ await processor.CloseAsync(cancellationToken)
...
                 }
             }
+            finally
+            {
+                processMessageEventArgs.MessageLockLostAsync -= MessageLockLostHandler;
+            }
+
+            async Task MessageLockLostHandler(MessageLockLostEventArgs lockLostArgs)
+            {
+                Logger.Info("Lost the lock while processing message. Cancelling the pipeline execution.", lockLostArgs.Exception);
```
johnsimons, line 367: I am unsure what is the benefit of adding the exception to the log in this log entry?

src/Transport/Receiving/MessagePump.cs
```diff
@@ -322,33 +325,56 @@ await processor.CloseAsync(cancellationToken)
...
+                try
+                {
+                    await processingTokenSource.CancelAsync().ConfigureAwait(false);
+                }
+                catch (ObjectDisposedException)
+                {
+                    // ignored
+                    Logger.Info("Lock lost handler executed but cancellation token source was already disposed.", lockLostArgs.Exception);
```
johnsimons, line 375: Should we bother logging this case?
Not much it can be done here.

src/Tests/Receiving/MessagePumpTests.cs
```diff
@@ -61,6 +61,38 @@ public async Task Should_not_process_message_when_lock_expired()
...
+                    await Task.Delay(Timeout.InfiniteTimeSpan, token);
+                },
+                (context, token) => Task.FromResult(ErrorHandleResult.Handled), CancellationToken.None);
+            await pump.StartReceive();
+
+            var receivedMessage = ServiceBusModelFactory.ServiceBusReceivedMessage(messageId: "SomeId", lockedUntil: DateTimeOffset.UtcNow.AddSeconds(60));
+
+            var fakeProcessor = fakeClient.Processors["receiveAddress"];
+            var processingTask = fakeProcessor.ProcessMessage(receivedMessage, fakeReceiver);
+            await fakeProcessor.RaiseMessageLockLost(receivedMessage, new MessageLockLostEventArgs(receivedMessage, new ServiceBusException("Lock Lost", ServiceBusFailureReason.MessageLockLost)));
```
johnsimons, line 87: Could there be a timing issue between this call and the one above?

johnsimons: Even if there is, I am unsure if it is avoidable.

johnsimons (approved):
Nice PR @danielmarbach.
I have added a few small comments for consideration.

src/Transport/Receiving/MessagePump.cs
SeanFeldman: If exception contains the message ID it helps. But, this message is not NSB message ID. And because physical message ID changes on delayed retries, it's less helpful during troubleshooting.

So I'd say maybe log NSB message ID.

src/Transport/Receiving/MessagePump.cs
danielmarbach: The problem is that this name is already taken by the pump token. I will rename it to `lockLostCancellationToken`

src/Transport/Receiving/MessagePump.cs
danielmarbach: The pump does enforce the message id to be set currently and throws otherwise. The message ID is already logged in a few places so I extended and aligned it here too.

src/Transport/Receiving/MessagePump.cs
danielmarbach: I think it is worthwhile having some logging in case we ever have to trouble shoot weird races. Should I go down to debug level?

src/Tests/Receiving/MessagePumpTests.cs
danielmarbach: Done

src/Transport/Receiving/MessagePump.cs
danielmarbach: Switched to debug for now

danielmarbach: I also double checked things with our logging guidance and unfortunately we almost always pass the exception to the logger methods and our logging abstraction has no `LevelFormat` overload that enables passing parameters as formats plus the exception. This makes things not useful for structured loggers :(

src/Transport/Receiving/MessagePump.cs
```diff
@@ -196,7 +196,7 @@ async Task OnProcessMessage(ProcessMessageEventArgs arg)
             }
             catch (Exception ex) when (ex.IsCausedBy(messageProcessingCancellationTokenSource.Token))
             {
-                Logger.Debug("Message processing canceled.", ex);
+                Logger.Debug($"Processing of the message with id '{messageId}' canceled.", ex);
```
johnsimons, line 199: Cancelled

danielmarbach: We standardized on American English

https://www.merriam-webster.com/grammar/canceled-or-cancelled

src/Transport/Receiving/MessagePump.cs
```diff
@@ -272,28 +272,31 @@ await processor.CloseAsync(cancellationToken)
             // args.CancellationToken is currently not used because the v8 version that supports cancellation was designed
             // to not flip the cancellation token until the very last moment in time when the stop token is flipped.
             var contextBag = new ContextBag();
+            using var lockLostCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(messageProcessingCancellationToken);
+            CancellationToken lockLostCancellationToken = lockLostCancellationTokenSource.Token;
```
PhilBastian, line 276: why not var?

PhilBastian (approved):

src/Transport/Receiving/MessagePump.cs
danielmarbach: Just to have more opportunities to asynchronously chat with you ;)

danielmarbach:
Marking it temporarily as a draft since there are a few edge cases that might need some more thinking

PR #1019, opened 2024-08-06
Fallback to individual sends when batching is not possible

danielmarbach:
After discussions, we have concluded that using the TryAdd capability of the SDK and fall back to using individual sends in cases when the message is too large is to best way to move forward for now instead of introducing a SendOption extension that allows to selectively opt out from batching per message.

The first `TryAdd` can only fail when the message is too large. Subsequent `TryAdd` attempt to add as much as possible until it returns false (either too large or batch full) then the current batch is sent, a new one is created and the message is attempted to add again. If the batch was full before, but the message fits in, it will be added. Otherwise, it will be pulled out for individual sends and batching continues (see comment in code)

PR #1023, opened 2024-08-08
Fallback to individual sends when batching is not possible

PhilBastian:
Backport of #1019 for the 4.1 branch


johnsimons (approved):
nicely ported

PR #1031, opened 2024-08-27
Add LRU array and update exception handling for `receiveOnly`

TravisNickels:
This PR is a follow up related to:

- https://github.com/Particular/NServiceBus.AmazonSQS/issues/1926
- https://github.com/Particular/NServiceBus.AmazonSQS/pull/1861

### ✔ What has been done

- Added an LRU (least recently used) cache from the [BitFaster Caching](https://github.com/bitfaster/BitFaster.Caching) library 
- When using `receiveOnly`, skip `onError()` and normal recoverability process if message lost lock exception is detected when trying to safely complete the message.
- If message lock lost exception is detected during the `onError()` the `messageId` will be added to the LRU
- If the message is found in the LRU, but is not able to remove the message from the queue it remains in the LRU cache.
 
### 💭 Assumptions

Applies only to `receiveOnly`

- Should try and maintain consistency with other transports that have implemented LRU caching (e.g., RabittMQ, SQS and ASQ). 
- Don't want to stop the pipeline of a `recieveOnly` receiver when the message lock has expired to ensure that the handler successfully finishes.
- The `messageId` then should be added to the LRU when the handler has finished, and a message lock lost exception has been thrown.  This will mark the message for deletion so that the handler can be skipped, and the message can be removed from the queue when it is encountered by the receiver instance again.
- Normal recoverability process should be skipped because it requires a message lock to be active to be able to delete the message from the queue and send a copy to the error queue.  However, since the message lock lost exception was thrown, and the instance no longer has access to the message, the normal recoverability process can't be executed.

### ⏰ Time Complexity

The big-Θ is linear with respect to N and D/M

big-Θ (N * (ceil (D / M))) 

N = number of receiver instances that could handle the message
D = the time it takes for the handler to finish
M = the max message lock duration and max auto lock renewal

Example: 

N = 2 (number of instances)
D = 33 seconds
M = 5 seconds

ceil(33/5) = ceil(6.6) = 7
big-Θ (2 * 7) = big-Θ (14)

This indicates that in the worst-case scenario (big-Θ) across all the receiver instances, the handler could possibly execute 14 times before the message is finally removed from the queue. 

In the best-case scenario (big-Ω) using the same example above, we could get lucky, and the same instance could handle the message, and the other instance never sees it.  However, since the handler still takes longer than the message lock duration, the best-case would still only be big-Ω(7).  

This means that it's still important to make sure that the handler execution time is less than or equal to the message lock duration and renewal to reduce big-Θ.

### DIFF
--- a/src/AcceptanceTests/NServiceBus.Transport.AzureServiceBus.AcceptanceTests.csproj
+++ b/src/AcceptanceTests/NServiceBus.Transport.AzureServiceBus.AcceptanceTests.csproj
@@ -13,6 +13,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.12.0" />
     <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.18.1" />
+    <PackageReference Include="BitFaster.Caching" Version="2.1.1" />
     <PackageReference Include="GitHubActionsTestLogger" Version="2.4.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.10.0" />
     <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="9.2.2" GeneratePathProperty="true" />
--- a/src/CommandLine/NServiceBus.Transport.AzureServiceBus.CommandLine.csproj
+++ b/src/CommandLine/NServiceBus.Transport.AzureServiceBus.CommandLine.csproj
@@ -12,6 +12,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.12.0" />
     <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.17.5" />
+    <PackageReference Include="BitFaster.Caching" Version="2.1.1" />
     <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="4.1.1" />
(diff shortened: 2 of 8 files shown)


danielmarbach:
Generally I'm a bit concerned about this change due to the following reasons

- The current version hides side effect occurring thins in a query named method
- The query named method is used in various places to compensate for effects that have a clear target from which they are caused, hiding the root cause and making the code harder to understand

I have [created a draft PR for discussion ](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1034)that shows some pseudocode and tests that show how things could be tackled. By moving the concrete scenarios closers to the complete and abandon it is possible to make targeted comments and log statements that make it clearer what's going on.

But even in that version, it becomes apparent that the current pump design might be suboptimal and might need to be revisited if we don't want to suffer through a series of bugs or behavior breaking changes by introducing this. I have a hunch it might be warranted to create some sort of receive strategy that is replaced depending on the transport transaction mode allowing to clearly separate the state tracking and error behavior for receive only vs sends with atomic receive

I think it is useful to closely investigate the current flow of complete, abandon and exception tracking and then align it closely on a case by case basis while using that knowledge to reconsider the logic flow of the pump

danielmarbach:
Replaced by https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1034

PR #1049, opened 2024-09-24
Inline dispatching of individual sends for messages that are too large to fit into a batch

### LINKED ISSUE #1048: Message too large for batch and then not sent
#### Description
Logging showed that:
"Message '4aae1ee3-7fca-432f-9a35-b1f100ae9abe' is too large for the batch '0' and will be sent individually to destination sbq-acc-ctb-shared-external-meteringworker."
But then nothing happened: the message is not sent.

#### Expected behavior
Message is sent and either the message is placed in the Azure Service Bus OR an exception is thrown if the message is too large for the queue.

#### Actual behavior
Nothing happens after reporting that it is too large for the batch.

#### Versions
NServiceBus 9.1.1 
Azure Service Bus transport 4.2.0

### Steps to reproduce

Send message that is larger than batch size (1024KB). Message size configured on the queue does not matter. 

### Relevant log output

```shell
On environment A:
"Message '4aae1ee3-7fca-432f-9a35-b1f100ae9abe' is too large for the batch '0' and will be sent individually to destination sbq-acc-ctb-shared-external-meteringworker."
and then nothing.
```

#### Workarounds
None.

#### Possible solutions
Looking at the code I suspect an issue in MessageDispatcher.
https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/Sending/MessageDispatcher.cs#L134

On line 134 a Task to execute DispatchBatchForDestination is added to dispatchTasks and passes the list messagesTooLargeToBeBatched into it.
Then on line 136 that list is iterated immediately, but I think the DispatchBatchForDestination may not have been executed then yet, so if a message is too large it will only be added to the list after the iteration on line 136, with an empty list, has already happened.

Could this be what is happening? I also see errors that the message is sent and then fails because the message is too large for the configured queue. Not sure how that can happen, as I would suspect it always loses the messages, so maybe something is missing in my logic.

danielmarbach:
Fixes #1048

[Previous test run of first commit before rebase that reproduced the problem](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/actions/runs/11017750455)

I think as a follow-up PR for master it might make sense to tone down slightly the overly clever trickery that fell apart when we introduced the collecting of too large messages, but I figured these structural changes should be done outside the realm of fixing the bug.

### DIFF
--- a/src/Tests/FakeSender.cs
+++ b/src/Tests/FakeSender.cs
@@ -27,27 +27,27 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests
             set => throw new NotSupportedException();
         }
 
-        public override ValueTask<ServiceBusMessageBatch> CreateMessageBatchAsync(CancellationToken cancellationToken = default)
+        public override async ValueTask<ServiceBusMessageBatch> CreateMessageBatchAsync(CancellationToken cancellationToken = default)
         {
             var batchMessageStore = new List<ServiceBusMessage>();
             ServiceBusMessageBatch serviceBusMessageBatch = ServiceBusModelFactory.ServiceBusMessageBatch(256 * 1024, batchMessageStore, tryAddCallback: TryAdd);
             batchToBackingStore.Add(serviceBusMessageBatch, batchMessageStore);
-            return new ValueTask<ServiceBusMessageBatch>(
-                serviceBusMessageBatch);
+            await Task.Yield();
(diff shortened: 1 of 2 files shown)


DavidBoike (approved):

PhilBastian:
needs a test to ensure someone doesn't see the Task.Yield's and revert the change thinking it's a mistake

PR #1052, opened 2024-09-25
Simplify dispatcher

danielmarbach:
Follow through on #1049 

This attempts to simplify the dispatcher a bit where possible to no longer rely on the synchronous path. This PR attempts to find a good balance between allocating the necessary task lists with the number of items needed because that is beneficial compared to doing `Task.WhenAll` with `Select` and `ToArray`

image

During the discussion with @andreasohlund and @PhilBastian we concluded the original problem would only really have been discovered if the following are true:

- Premium namespace (we don't use a premium namespace yet in our CI/CD environment)
- The underlying SDK cannot yet have cached the batch size information of the underlying AMQP link (will make the `CreateMessageBatchAsync` yield) or the large message has to be put into a batch that is sent after the first batch was dispatched

Given that we have not even the premium namespace in place currently, simplifying the dispatcher slightly to avoid such an error seems to be warranted.

I have also removed the yielding, since that was just a way to reproduce the previous problem. With the new structure this is no longer necessary.

### DIFF
--- a/src/Transport/Sending/MessageDispatcher.cs
+++ b/src/Transport/Sending/MessageDispatcher.cs
@@ -44,7 +44,6 @@ namespace NServiceBus.Transport.AzureServiceBus
             Dictionary<string, List<IOutgoingTransportOperation>>? defaultOperationsPerDestination = null;
             var numberOfDefaultOperations = 0;
             var numberOfIsolatedOperations = 0;
-            var numberOfDefaultOperationDestinations = 0;
 
             foreach (var operation in transportOperations)
             {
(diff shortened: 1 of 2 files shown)


src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -59,8 +58,6 @@ public async Task Dispatch(TransportOperations outgoingMessages, TransportTransa
                         if (!defaultOperationsPerDestination.ContainsKey(destination))
                         {
                             defaultOperationsPerDestination[destination] = [operation];
-                            // because we batch only the number of destinations are relevant
```
danielmarbach, line 62: Getting rid of this here also reduces the cognitive overhead because this code doesn't have knowledge anymore about the inner workings of the batching path.

DavidBoike (approved):

PhilBastian (approved):

andreasohlund (approved):

PR #1056, opened 2024-09-30
Fix thread-safe comparison used to optimize circuit breaker's Success method

DavidBoike:
The circuit breaker's Success() method must be as close to a no-op as possible on the happy path, but when disarming in a multi-threaded environment still needs to run the disarm logic only once.

Previously, while querying values was done in a thread-safe way, a thread on the Failure method could pause while a Success() call ran to entirety, setting the failure count to 0, and ensuring that further calls to Success() would see a failure count of 0 and take no action, even though the arming action had been executed most recently.

This resulted in a scenario where no number of Success calls could undo the armedAction, unless another Failure call went first.

### DIFF
--- a/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
+++ b/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
@@ -29,6 +29,7 @@
                 return;
             }
 
+            Volatile.Write(ref failureCount, 0);
             timer.Change(Timeout.Infinite, Timeout.Infinite);
             Logger.InfoFormat("The circuit breaker for {0} is now disarmed", name);
             disarmedAction();
--- a/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
+++ b/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
@@ -23,9 +23,8 @@
 
         public void Success()
         {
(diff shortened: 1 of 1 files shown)


src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -23,9 +23,8 @@ public RepeatedFailuresOverTimeCircuitBreaker(string name, TimeSpan timeToWaitBe
 
         public void Success()
         {
-            var oldValue = Interlocked.Exchange(ref failureCount, 0);
-
-            if (oldValue == 0)
+            // If the failure count was already zero, replace it with zero (no change) and then return original
+            if (Interlocked.CompareExchange(ref failureCount, 0, 0) == 0)
```
awright18, line 27: This seems like it would solve both problems.

//if there have been no failures previously then return
if(Interlocked.Read(ref failureCount) == 0 )
{
	return;
}
else //if there were previous failures then it's ok to disarm now.
{
   if(Interlocked.Read(ref failureCount) > 0) //not sure if this check is necessary or not.
   {
	  Interlocked.Exchange(ref failureCount, 0);
   }
}

danielmarbach:
Here is a hopefully simplified version https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1057

DavidBoike:
Closing in favor of https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1057

PR #1057, opened 2024-09-30
Address races in the repeated failure circuit breaker by introducing explicit state transitions

danielmarbach:
This implementation simplifies the implementation of the circuit breaker by no longer incrementing the failure count. The circuit breaker doesn't really care about the number of failures. It cares about tracking the potential state transitions between `Armed`, `Disarmed` and `Triggered`.

By focusing the implementation on these state transitions, the complex interaction between the `failureCount` and the `triggered` flag can be removed and replaced by a state that has meaningful names.

### DIFF
--- a/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
+++ b/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
@@ -45,7 +45,7 @@
             {
                 armedAction();
                 _ = timer.Change(timeToWaitBeforeTriggering, NoPeriodicTriggering);
-                Logger.WarnFormat("The circuit breaker for {0} is now in the armed state", name);
+                Logger.WarnFormat("The circuit breaker for {0} is now in the armed state due to {1}", name, exception);
             }
 
             // If the circuit breaker has been triggered, wait for 10 seconds before proceeding to prevent flooding the logs and hammering the ServiceBus
@@ -61,7 +61,7 @@
                 return;
             }
 
-            Logger.WarnFormat("The circuit breaker for {0} will now be triggered", name);
+            Logger.WarnFormat("The circuit breaker for {0} will now be triggered with exception {1}", name, lastException);
             triggerAction(lastException);
         }
 
--- a/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
+++ b/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
@@ -1,4 +1,4 @@
-﻿namespace NServiceBus.Transport.AzureServiceBus
+namespace NServiceBus.Transport.AzureServiceBus
 {
     using System;
     using System.Threading;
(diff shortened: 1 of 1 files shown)


src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -23,54 +23,49 @@ public RepeatedFailuresOverTimeCircuitBreaker(string name, TimeSpan timeToWaitBe
...
         {
-            lastException = exception;
-            var newValue = Interlocked.Increment(ref failureCount);
+            _ = Interlocked.Exchange(ref lastException, exception);
+
+            // Atomically set state to Armed if it was previously Disarmed
+            var previousState = Interlocked.CompareExchange(ref circuitBreakerState, Armed, Disarmed);
 
-            if (newValue == 1)
+            if (previousState == Disarmed)
```
awright18, line 44:
```suggestion
            if (Interlocked.CompareExchange(ref circuitBreakerState, Armed, Disarmed)== Disarmed)
```

src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -23,54 +23,49 @@ public RepeatedFailuresOverTimeCircuitBreaker(string name, TimeSpan timeToWaitBe
...
         }
 
         public Task Failure(Exception exception, CancellationToken cancellationToken = default)
         {
-            lastException = exception;
-            var newValue = Interlocked.Increment(ref failureCount);
+            _ = Interlocked.Exchange(ref lastException, exception);
+
+            // Atomically set state to Armed if it was previously Disarmed
+            var previousState = Interlocked.CompareExchange(ref circuitBreakerState, Armed, Disarmed);
```
awright18, line 42:
```suggestion
```

src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -23,54 +23,49 @@ public RepeatedFailuresOverTimeCircuitBreaker(string name, TimeSpan timeToWaitBe
 
         public void Success()
         {
-            var oldValue = Interlocked.Exchange(ref failureCount, 0);
+            var previousState = Interlocked.CompareExchange(ref circuitBreakerState, Disarmed, Armed);
 
-            if (oldValue == 0)
+            // If the circuit breaker was Armed or triggered before, disarm it
+            if (previousState == Armed || Interlocked.CompareExchange(ref circuitBreakerState, Disarmed, Triggered) == Triggered)
```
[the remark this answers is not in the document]
awright18: probably even nicer would be methods IsCircuitBreakerArmed() and IsCircuitBreakerTiggered() and those being called in the if block, but maybe that indirection would have a performance impact.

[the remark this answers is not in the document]
DavidBoike: This code is hard enough to read, I like having a label on the result of the CompareExchange. I trust the compiler to optimize it away anyway.

src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
DavidBoike: Plus, down here, `previousState` is used again on line 52, and it's good to have them both aligned for readability.

DavidBoike (approved):

src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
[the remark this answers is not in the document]
danielmarbach: Here it is definitely not needed. It could be inlined. I'm good either way.

src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
danielmarbach: Yeah, otherwise you would have to volatile read the `circuitBreakerState` again

awright18 (approved):

PR #1066, opened 2024-10-01
Use locks to serialize arm/disarm actions

DavidBoike:
Serializing the arm/disarm actions is not enough. The state transition preceding the arm/disarm action must be captured in the lock as well, otherwise the state transitions could happen properly but then the arm/disarm actions could be executed by the processor in the wrong order, leaving the logic of the circuit breaker itself intact but the consequences of the arm/disarm actions in an incorrect state.

I also added xdoc as a way to document that the arm/disarm actions are called within a lock, and to maybe pave the way for this to be used as a shared component.

### DIFF
--- a/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
+++ b/src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
@@ -69,7 +69,15 @@ namespace NServiceBus.Transport.AzureServiceBus
 
                 _ = timer.Change(Timeout.Infinite, Timeout.Infinite);
                 Logger.InfoFormat("The circuit breaker for {0} is now disarmed", name);
-                disarmedAction();
+                try
+                {
+                    disarmedAction();
+                }
+                catch (Exception ex)
+                {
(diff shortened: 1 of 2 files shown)


danielmarbach (approved):

src/Transport/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -1,79 +1,153 @@
...
         public Task Failure(Exception exception, CancellationToken cancellationToken = default)
         {
+            // Atomically store the exception that caused the circuit breaker to trip
             _ = Interlocked.Exchange(ref lastException, exception);
 
-            // Atomically set state to Armed if it was previously Disarmed
-            var previousState = Interlocked.CompareExchange(ref circuitBreakerState, Armed, Disarmed);
+            // Check the status of the circuit breaker, exiting early outside the lock if already armed or triggered
+            var previousState = Volatile.Read(ref circuitBreakerState);
+            if (previousState != Disarmed)
```
danielmarbach, line 87: Was wondering if an  `is Armed or Triggered` would improve the readability here, but then it is also more instructions

PR #1071, opened 2024-10-10
Allow overriding native MessageId, ContentType and CorrelationID

ramonsmits:
Alternatives:

1. Have a generic extension point `Action<TransportOperation, ServiceBusMessage>` that any customizations on the native message type and use this from the ******** connector.
2. Add extensions methods to transport operations to have type safe APIs to set these values explicitly.

### DIFF
--- a/src/Transport/Sending/OutgoingMessageExtensions.cs
+++ b/src/Transport/Sending/OutgoingMessageExtensions.cs
@@ -7,8 +7,11 @@
 
     static class OutgoingMessageExtensions
     {
-        public static ServiceBusMessage ToAzureServiceBusMessage(this OutgoingMessage outgoingMessage, DispatchProperties dispatchProperties, string incomingQueuePartitionKey)
+        public static ServiceBusMessage ToAzureServiceBusMessage(this IOutgoingTransportOperation outgoingTransportOperation, string incomingQueuePartitionKey)
         {
+            var outgoingMessage = outgoingTransportOperation.Message;
+            var message = new ServiceBusMessage(outgoingMessage.Body);
+            var dispatchProperties = outgoingTransportOperation.Properties;
 
             ApplyMessageId(message, outgoingTransportOperation);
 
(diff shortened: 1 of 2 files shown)


src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -51,17 +47,31 @@ static void ApplyDeliveryConstraints(ServiceBusMessage message, DispatchProperti
...
         }
 
-        static void ApplyCorrelationId(ServiceBusMessage message, Dictionary<string, string> headers)
+        static void ApplyCorrelationId(ServiceBusMessage message, IOutgoingTransportOperation outgoingTransportOperation)
         {
-            if (headers.TryGetValue(Headers.CorrelationId, out var correlationId))
+            var properties = outgoingTransportOperation.Properties;
+            var headers = outgoingTransportOperation.Message.Headers;
+
+            if (properties.TryGetValue(Headers.CorrelationId, out var correlationId))
```
danielmarbach, line 55: As far as I recall, the properties get serialized into the outbox. It might be worthwhile considering removing those temporary properties after reading them or add a comment why they need to be preserved

danielmarbach: Quick question. Technically, already today, it is possible to add a customization delegate that allows you to customize the entire native message created. [Any reason this can't be used](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/3dd31681a9e4fc42c5bcd549ed9f56b944cbf9e1/src/Transport/Sending/OutgoingTransportOperationExtensions.cs#L26)?

danielmarbach: And yes, it seems the customizer should also be removed from the properties dictionary because I doubt we can every serialize action delegates into outbox records.

ramonsmits: 1. I didn't know that `TransportTransaction` was itself a `ContextBag` 🤯
2. Looked at the type you linked but I haven't figured out how that could be used as so far all invoked types I see are internal

So, I need additional research to validate if this could be used for the ******* connector.

PR #1072, opened 2024-10-10
Ability to not generate legacy `NServiceBus.Transport.Encoding` header

ramonsmits:
Add operation extension to not generate legacy `NServiceBus.Transport.Encoding` header in native custom properties / ApplicationProperties

### DIFF
--- a/src/Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
+++ b/src/Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
@@ -6,6 +6,11 @@ namespace NServiceBus
     {
         public AzureServiceBusTransport(string connectionString) { }
         public AzureServiceBusTransport(string fullyQualifiedNamespace, Azure.Core.TokenCredential tokenCredential) { }
+        [System.Obsolete("Next versions of the transport will by default no longer send the transport encod" +
+            "ing header for wire compatibility, requiring an opt-in for the header to be sent" +
+            ". Will be treated as an error from version 5.0.0. Will be removed in version 6.0" +
+            ".0.", false)]
+        public bool DoNotSendTransportEncodingHeader { get; set; }
         public bool EnablePartitioning { get; set; }
         public int EntityMaximumSize { get; set; }
         public System.TimeSpan? MaxAutoLockRenewalDuration { get; set; }
@@ -32,6 +37,11 @@ namespace NServiceBus
         public static NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> ConnectionString(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, string connectionString) { }
         public static NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> CustomRetryPolicy(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, Azure.Messaging.ServiceBus.ServiceBusRetryOptions retryPolicy) { }
         public static NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> CustomTokenCredential(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, string fullyQualifiedNamespace, Azure.Core.TokenCredential tokenCredential) { }
+        [System.Obsolete("Next versions of the transport will by default no longer send the transport encod" +
+            "ing header for wire compatibility, requiring an opt-in for the header to be sent" +
+            ". Will be treated as an error from version 5.0.0. Will be removed in version 6.0" +
+            ".0.", false)]
+        public static NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> DoNotSendTransportEncodingHeader(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> EnablePartitioning(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> EntityMaximumSize(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, int maximumSizeInGB) { }
         public static NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> MaxAutoLockRenewalDuration(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, System.TimeSpan maximumAutoLockRenewalDuration) { }
--- a/src/Transport/AzureServiceBusTransport.cs
+++ b/src/Transport/AzureServiceBusTransport.cs
@@ -329,6 +329,14 @@
         }
         IWebProxy webProxy;
 
+        /// <summary>
+        /// When set will not add `NServiceBus.Transport.Encoding` header for wire compatibility with NServiceBus.AzureServiceBus. The default value is <c>false</c>.
+        /// </summary>
+        [ObsoleteEx(Message = "Next versions of the transport will by default no longer send the transport encoding header for wire compatibility, requiring an opt-in for the header to be sent.",
+            TreatAsErrorFromVersion = "5",
+            RemoveInVersion = "6")]
+        public bool DoNotSendTransportEncodingHeader { get; set; }
+
         internal string ConnectionString { get; set; }
 
         internal string FullyQualifiedNamespace { get; set; }
--- a/src/Transport/AzureServiceBusTransportSettingsExtensions.cs
+++ b/src/Transport/AzureServiceBusTransportSettingsExtensions.cs
@@ -229,7 +229,8 @@
         /// Overrides the default maximum duration within which the lock will be renewed automatically. This
         /// value should be greater than the longest message lock duration.
         /// </summary>
-        /// <value>The maximum duration during which message locks are automatically renewed. The default value is 5 minutes.</value>
+        /// <param name="transportExtensions"></param>
+        /// <param name="maximumAutoLockRenewalDuration">The maximum duration during which message locks are automatically renewed. The default value is 5 minutes.</param>
         /// <remarks>The message renew can continue for sometime in the background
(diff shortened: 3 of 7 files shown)


src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -7,16 +7,21 @@
...
+            var outgoingMessage = outgoingTransportOperation.Message;
             var message = new ServiceBusMessage(outgoingMessage.Body)
             {
                 // Cannot re-use MessageId to be compatible with ASB transport that could have native de-dup enabled
                 MessageId = Guid.NewGuid().ToString()
             };
 
-            // The value needs to be "application/octect-stream" and not "application/octet-stream" for interop with ASB transport
-            message.ApplicationProperties[TransportMessageHeaders.TransportEncoding] = "application/octect-stream";
+            var disableLegacyHeaders = outgoingTransportOperation.Properties.ContainsKey(TransportOperationExt.DisableLegacyHeadersKey);
```
danielmarbach, line 19: If I recall correctly, those properties get serialized into the outbox too. It might be worthwhile adding a comment here to indicate why the property is not removed (because it needs to be preserved even with the outbox if I understand this correctly).

From a more general perspective I wonder if it would make sense to actually moved this up to the transport level configuration. For all users that do not need to interop the old transport this header is unnecessary. By moving it to the transport level we can make it opt-out, later opt-in and eventually deprecate it from the transport entirely.

This seems to be the more future-proof way of handling this change instead of introducing yet another configuration layer deep down on the transport operation level.

ramonsmits: @danielmarbach I like your reasoning about uplifting this and not have it per operation but as an endpoint configuration. Here it would be AND also on the top level configuration as the ***** connector uses the transport seam.

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -333,5 +333,14 @@ public IWebProxy WebProxy
 
         internal string FullyQualifiedNamespace { get; set; }
         internal TokenCredential TokenCredential { get; set; }
+
+        /// <summary>
+        /// Disable legacy headers for wire compatibility with NServiceBus.AzureServiceBus
+        /// </summary>
+        public void DisableLegacyHeaders()
```
danielmarbach, line 340: FYI when you look at the transport definition we opted for a more property-based model. Only the legacy shim extensions that make the transport compatible with `UseTransport<AzureServiceBusTransport>` uses methods.

ramonsmits: > FYI when you look at the transport definition we opted for a more property-based model. Only the legacy shim extensions that make the transport compatible with `UseTransport<AzureServiceBusTransport>` uses methods.

I opted for a method because of "Disable". Feels more natural as an "opt-out" API compared to doing:

```
transport.DisableLegacyHeaders = true
```

Alternative could  be:

```
transport.LegacyHeaders = false
```

Now that I think a bit longer about this API it likely would be better to mention the compabitibility in the name:

```
transport.DisableLegacyTransportCompatibility()
```

danielmarbach: Look at other transports

https://github.com/Particular/NServiceBus.AmazonSQS/blob/master/src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
https://github.com/Particular/NServiceBus.RabbitMQ/blob/master/src/NServiceBus.Transport.RabbitMQ/RabbitMQTransport.cs
https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.Transport.SqlServer/SqlServerTransport.cs
https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.Transport.PostgreSql/PostgreSqlTransport.cs
https://github.com/Particular/NServiceBus.Transport.Msmq/blob/master/src/NServiceBus.Transport.Msmq/MsmqTransport.cs

etc. They all use properties to express the simple configuration flags and name the properties according to the intent (some of the internal ones even contain `Disable`). The only different API is the one on Rabbit with the Add semantic where it makes sense even though it could also have been expressed with a more complex type on a property.

Opting for a different approach based on the above justification doesn't seem warranted because it seems to fall into the category of "I like it better this way".

danielmarbach: Furthermore, [considering this guidance](https://github.com/Particular/Platform/blob/main/documentation/transports.md) (copying and pasting for reference)

> The shim API methods are currently marked with a [[PreObsolete] attribute](https://github.com/Particular/Platform/blob/main/documentation/pre-obsolete-attribute.md) - this is in preparation to [making the API obsolete](https://github.com/Particular/Platform/blob/main/guidelines/breaking-changes.md#deprecating-an-api). However, **since the API is not yet marked as obsolete, and hence users are not seeing warnings about using this API, any new transport settings should always be added to both**. This will ensure that users of both implementation methods are able to make use of new features without rewriting their code.
> 
> Once we decide to retire the shim API and its methods are marked as obsolete, we can then stop modifying it and have new settings only added to the new API.

It means the missing shim API method would clash anyway.

danielmarbach: Why not `DoNotSendTransportEncodingHeaders`. The more general the name is, the more likely it will collapse with some other headers that we might need to deprecate. So, exposing a very concrete property for the specific header we are no longer sending seems to be better, no? Because the term "Legacy" is very broad and might chance over the multiple version of the transport we have to release until we can finally deprecate the setting, having such a broad term seems almost dangerous.

ramonsmits: That header is only used by ASBL. Users might otherwise not know why they could invoke that but that still is in the property docs. I like the naming compared to "Disabled" so adopted that.

danielmarbach: DoNotSendTransportEncodingHeaderForBackwardsCompatibilityWithPreviousTransportVersions? :dancing_men:

irinascurtu (approved):

src/Transport/AzureServiceBusTransport.cs
ramonsmits: I applied your suggestion

SzymonPobiega (approved):

tmasternak (approved):

src/Transport/Sending/OutgoingMessageExtensions.cs
ramonsmits: Applied on PR

danielmarbach:
@ramonsmits Reviewing...

danielmarbach:
@ramonsmits I added https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1082. My understanding was that with the transport level approach, the per operation level approach is no longer required

As highlighted in the comment that was resolved this is also missing the legacy shim extensions

src/Transport/AzureServiceBusTransportSettingsExtensions.cs
```diff
@@ -229,7 +229,8 @@ public static TransportExtensions<AzureServiceBusTransport> CustomTokenCredentia
         /// Overrides the default maximum duration within which the lock will be renewed automatically. This
         /// value should be greater than the longest message lock duration.
         /// </summary>
-        /// <value>The maximum duration during which message locks are automatically renewed. The default value is 5 minutes.</value>
+        /// <param name="transportExtensions"></param>
+        /// <param name="maximumAutoLockRenewalDuration">he maximum duration during which message locks are automatically renewed. The default value is 5 minutes..</param>
```
danielmarbach, line 233:
```suggestion
        /// <param name="maximumAutoLockRenewalDuration">The maximum duration during which message locks are automatically renewed. The default value is 5 minutes.</param>
```

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -329,6 +329,14 @@ public IWebProxy WebProxy
         }
         IWebProxy webProxy;
 
+        /// <summary>
+        /// When set will not add `NServiceBus.Transport.Encoding` header for wire compatibility with NServiceBus.AzureServiceBus. The default value is <c>false</c>.
+        /// </summary>
+        [ObsoleteEx(Message = "Next versions of the transport will be default no longer send the transport encoding header for wire compatibility by default requiring an opt-in for the header to be sent.",
```
WilliamBZA, line 335:
```suggestion
        [ObsoleteEx(Message = "Next versions of the transport will by default no longer send the transport encoding header for wire compatibility, requiring an opt-in for the header to be sent.",
```

src/Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -6,6 +6,11 @@ namespace NServiceBus
     {
         public AzureServiceBusTransport(string connectionString) { }
         public AzureServiceBusTransport(string fullyQualifiedNamespace, Azure.Core.TokenCredential tokenCredential) { }
+        [System.Obsolete("Next versions of the transport will be default no longer send the transport encod" +
+            "ing header for wire compatibility by default requiring an opt-in for the header " +
+            "to be sent. Will be treated as an error from version 5.0.0. Will be removed in v" +
+            "ersion 6.0.0.", false)]
+        public bool DoNotSendTransportEncodingHeader { get; set; }
```
WilliamBZA, line 13: This would need to be updated too based on if the comment below is included

src/Transport/AzureServiceBusTransportSettingsExtensions.cs
```diff
@@ -241,5 +242,18 @@ public static TransportExtensions<AzureServiceBusTransport> MaxAutoLockRenewalDu
             transportExtensions.Transport.MaxAutoLockRenewalDuration = maximumAutoLockRenewalDuration;
             return transportExtensions;
         }
+
+        /// <summary>
+        /// When set will not add `NServiceBus.Transport.Encoding` header for wire compatibility with NServiceBus.AzureServiceBus. The default value is <c>false</c>.
+        /// </summary>
+        /// <param name="transportExtensions"></param>
+        [ObsoleteEx(Message = "Next versions of the transport will be default no longer send the transport encoding header for wire compatibility by default requiring an opt-in for the header to be sent.",
```
WilliamBZA, line 250:
```suggestion
        [ObsoleteEx(Message = "Next versions of the transport will by default no longer send the transport encoding header for wire compatibility, requiring an opt-in for the header to be sent.",
```

WilliamBZA (commented):
Minor tweaks

tmasternak (approved):

PR #1082, opened 2024-10-23
Disable legacy headers review

danielmarbach:


danielmarbach:
This is still missing

> The new feature is introduced
>     It is not enabled by default
>     Add an API to enable the feature
>         Deprecate the API by marking it as obsolete using an [ObsoleteEx attribute](https://github.com/Fody/Obsolete), treated as a warning in this version, and to be removed in the next major version.
>     Create an issue in the public repository to turn the feature on by default in the next major. Assign it to the appropriate milestone (create it if it doesn't exist).

which I can add later

danielmarbach:
Done

danielmarbach:
Raised https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/1083

PR #1085, opened 2024-10-24
Allow customizing outgoing native ServiceBusMessage

ramonsmits:
Replaces:

- https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1071

Added the ability to customize then native message type just before it is dispatched to Azure Service Bus and gets access to [IOutgoingTransportOperation](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Transports/IOutgoingTransportOperation.cs).

```c#
var transport = new AzureServiceBusTransport(connectionstring);
//...
transport.OutgoingNativeMessageCustomization = (outgoingoperation, servicebusmessage) => { servicebusmessage.ContentType = "text/xml" };
```

or

```c#
public Task MyCustomizer(IOutgoingTransportOperation message, ServiceBusMessage nativeMessage)
{
}

var transport = new AzureServiceBusTransport(connectionstring);
//...
transport.OutgoingNativeMessageCustomization = MyCustomizer;
```

### DIFF
--- a/src/AcceptanceTests/When_customizing_outgoing_messages.cs
+++ b/src/AcceptanceTests/When_customizing_outgoing_messages.cs
@@ -34,9 +34,8 @@ namespace NServiceBus.Transport.RabbitMQ.AcceptanceTests
             public Receiver() =>
                 EndpointSetup<DefaultServer>(endpointConfiguration =>
                 {
-                    var t = (AzureServiceBusTransport)endpointConfiguration.ConfigureTransport();
-                    t.OutgoingNativeMessageCustomization =
-                        (_, message) => message.Subject = TestSubject;
+                    var transport = endpointConfiguration.ConfigureTransport<AzureServiceBusTransport>();
+                    transport.OutgoingNativeMessageCustomization = (_, message) => message.Subject = TestSubject;
                 });
 
             class MyEventHandler(Context testContext) : IHandleMessages<Message>
--- a/src/AcceptanceTests/When_customizing_outgoing_messages.cs
+++ b/src/AcceptanceTests/When_customizing_outgoing_messages.cs
@@ -1,0 +1,70 @@
+namespace NServiceBus.Transport.RabbitMQ.AcceptanceTests
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Azure.Messaging.ServiceBus;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+
+    class When_customizing_outgoing_messages : NServiceBusAcceptanceTest
+    {
+        const string TestSubject = "0192c3ad-8ab2-77a0-8a92-2be53f062e06";
+
+        [Test]
+        public async Task Should_receive_custom_set_value()
+        {
+            var scenario = await Scenario.Define<Context>()
+                .WithEndpoint<Receiver>(b => b.When((bus, c) => bus.SendLocal(new Message())))
+                .Done(c => c.MessageReceived)
+                .Run();
+
+            Assert.That(scenario.ReceivedMessage.Subject, Is.EqualTo(TestSubject));
+        }
+
+        public class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(endpointConfiguration =>
+                {
+                    var t = (AzureServiceBusTransport)endpointConfiguration.ConfigureTransport();
+                    t.OutgoingNativeMessageCustomization =
+                        (operation, message) =>
+                        {
+                            message.Subject = TestSubject;
+                        };
+                });
+            }
+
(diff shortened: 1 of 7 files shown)


irinascurtu (approved):

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -333,5 +333,10 @@ public IWebProxy WebProxy
 
         internal string FullyQualifiedNamespace { get; set; }
         internal TokenCredential TokenCredential { get; set; }
+
+        /// <summary>
+        /// Register call that allows to customize the native ServiceBusMessage just before it is dispatched to the Azure Service Bus SDK client
+        /// </summary>
+        public NativeMessageCustomizerCallback NativeMessageCustomization { get; set; }
```
danielmarbach, line 340: Could we name this containing the word "outgoing" or anything the like to make it clear what the purpose is? Because technically, we could also have an incoming customizer.

Also what is the benefit of having a custom delegate over Action delegate? Is it the ability to name things? Could we maybe also promote this approach if we stick to it to our design principles in platform? I'm aware we have done the same for the OnMessage and OnError delegates which seem to be special cases due to the complexity and low level embedding into the transport. Given this is user facing it seems applying the framework guidelines would be a sensible thing to do in my opinion

>  DO use the new Func<...>, Action<...>, or Expression<...> types instead of custom delegates, when defining APIs with callbacks.

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/events-and-callbacks

src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -20,11 +20,17 @@ class MessageDispatcher : IMessageDispatcher
...
-        public MessageDispatcher(MessageSenderRegistry messageSenderRegistry, string topicName)
+        public MessageDispatcher(
+            MessageSenderRegistry messageSenderRegistry,
+            string topicName,
+            NativeMessageCustomizerCallback? customizerCallback = null
+        )
         {
             this.messageSenderRegistry = messageSenderRegistry;
             this.topicName = topicName;
+            this.customizerCallback = customizerCallback;
```
danielmarbach, line 33: By assigning a static empty version when null you remove the null check on the hot path

src/Transport/AzureServiceBusTransport.cs
SeanFeldman: To echo the name aspect, "customizer" and "callback" together read somewhat redundant. As a user I'd understand it's a hook to manipulate the dispatched message.

src/Transport/Sending/MessageDispatcher.cs
ramonsmits: Applied

src/AcceptanceTests/When_customizing_outgoing_messages.cs
```diff
@@ -0,0 +1,71 @@
...
+                Context testContext;
+
+                public MyEventHandler(Context testContext)
+                {
+                    this.testContext = testContext;
+                }
+
+                public Task Handle(Message message, IMessageHandlerContext context)
+                {
+                    testContext.ReceivedMessage = context.Extensions.Get<ServiceBusReceivedMessage>();
```
danielmarbach, line 52: Funny enough the application properties are headers. So you could simply check extract the context.Headers here in case you want. Assuming that would it better to set something like subject?

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -333,5 +333,10 @@ public IWebProxy WebProxy
 
         internal string FullyQualifiedNamespace { get; set; }
         internal TokenCredential TokenCredential { get; set; }
+
+        /// <summary>
+        /// Register call that allows to customize the native ServiceBusMessage just before it is dispatched to the Azure Service Bus SDK client
+        /// </summary>
```
danielmarbach, line 339: How about something like this?

```suggestion
        /// <summary>
        /// Gets or sets the action that allows customization of the native <see cref="ServiceBusMessage"/> 
        /// just before it is dispatched to the Azure Service Bus SDK client. 
        /// </summary>
        /// <remarks>
        /// <para>
        /// This customization is applied after any configured transport customizations, meaning that 
        /// any changes made here may override or conflict with previous transport-level adjustments. 
        /// Exercise caution, as modifying the message at this stage can lead to unintended behavior 
        /// downstream if the message structure or properties are altered in ways that do not align 
        /// with expectations elsewhere in the system.
        /// </para>
        /// </remarks>
```

src/AcceptanceTests/When_customizing_outgoing_messages.cs
ramonsmits: I didn't touch other properties to avoid conflict with transport used properties. Specifically because it now uses an AT but I'll test with subject. Didn't know that property yet.

ramonsmits: Applied on branch

danielmarbach (approved):

PR #1094, opened 2024-11-07
Stop pumps on shutdown

danielmarbach:
This aligns the transport to stop the pumps on transport shutdown and at the same time makes sure the pumps stop method can be called multiple times. 

This is part of addressing some low-hanging fruit to align all transports for the transport seam usage scenarios to make sure pumps are stopped when the transport shuts down. 

I have tried to write a transport test for it to enforce it eventually for all transport but couldn't find a way to do that without introducing flags on the transport interface and using fancy techniques like default members which seams overkill

### DIFF
--- a/src/Transport/AzureServiceBusTransportInfrastructure.cs
+++ b/src/Transport/AzureServiceBusTransportInfrastructure.cs
@@ -85,6 +85,9 @@
 
         public override async Task Shutdown(CancellationToken cancellationToken = default)
         {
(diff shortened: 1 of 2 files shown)


SeanFeldman (approved):

SeanFeldman:
> I have tried to write a transport test for it to enforce it eventually for all transport but couldn't find a way to do that without introducing flags on the transport interface and using fancy techniques like default members which seams overkill

Could just say "Testing will be performed using PUnit".

andreasohlund (approved):

PR #759, opened 2023-01-19
Dual target command line

danielmarbach:
Fixes #758

PR #1146, opened 2025-01-31
Update the command line tool to not use .Net 7

tamararivera:
The absence of .Net 7 on the Github runner is breaking the build pipeline. 
.Net 7 was added on https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/759
This change makes the command line tool single target, but sets the roll forward behavior to `Major` and adds the target framework to the command that the tests execute.

### DIFF
--- a/src/CommandLine/NServiceBus.Transport.AzureServiceBus.CommandLine.csproj
+++ b/src/CommandLine/NServiceBus.Transport.AzureServiceBus.CommandLine.csproj
@@ -7,6 +7,7 @@
     <PackAsTool>True</PackAsTool>
     <Description>.NET Core global tool to manage Azure Service Bus entities for NServiceBus endpoints</Description>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <RollForward>Major</RollForward>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


src/CommandLine/NServiceBus.Transport.AzureServiceBus.CommandLine.csproj
```diff
@@ -1,12 +1,13 @@
...
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
+    <TargetFramework>net6.0</TargetFramework>
     <OutputType>Exe</OutputType>
     <ToolCommandName>asb-transport</ToolCommandName>
     <PackAsTool>True</PackAsTool>
     <Description>.NET Core global tool to manage Azure Service Bus entities for NServiceBus endpoints</Description>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <RollForward>Major</RollForward>
```
tamararivera, line 10: I wasn't sure on what should I pick after reading [this](https://learn.microsoft.com/en-us/dotnet/core/versions/selection#values), so I decided to go with Major

bording (approved):

PR #1155, opened 2025-02-10
Adding support to new topology to the command line tool

SzymonPobiega:
Adds support to new topology to the command line tool:
 - `topology` is a new required parameter for the `endpoint` command with values either `migration` or `topic-per-event`
 - `subscribe` and `unsubscribe` use topic-per-event topology
 - `subscribe-non-migrated` and `unsubscribe-non-migrated` use the old topology

Note: it seems impossible to use single "subscribe" commands with a topology parameter because, depending on the topology, the arguments for the commands are different (not only options).

### DIFF
--- a/src/CommandLine/Program.cs
+++ b/src/CommandLine/Program.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.ComponentModel.DataAnnotations;
+    using System.Diagnostics;
     using Azure.Messaging.ServiceBus;
     using McMaster.Extensions.CommandLineUtils;
 
@@ -56,7 +57,11 @@
                     createCommand.AddOption(fullyQualifiedNamespace);
                     createCommand.AddOption(size);
                     createCommand.AddOption(partitioning);
-                    var topicName = createCommand.Option("-t|--topic", "Topic name (defaults to 'bundle-1')", CommandOptionType.SingleValue);
+
+                    var topology = createCommand.Option("-t|--topology", "Topology to use", CommandOptionType.SingleValue)
+                        .IsRequired()
+                        .Accepts(v => v.Values("migration", "topic-per-event"));
+                    var topicName = createCommand.Option("-st|--topic", "Topic name (defaults to 'bundle-1')", CommandOptionType.SingleValue);
                     var topicToPublishTo = createCommand.Option("-tp|--topic-to-publish-to", "The topic name to publish to", CommandOptionType.SingleValue);
                     var topicToSubscribeOn = createCommand.Option("-ts|--topic-to-subscribe-on", "The topic name to subscribe on", CommandOptionType.SingleValue);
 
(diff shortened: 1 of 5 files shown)


SzymonPobiega:
@danielmarbach this is now in the form that I think is ready for a review.

danielmarbach:
I'm wondering if it wouldn't be better to have a more drastic approach with the tooling by moving everything migration related underneath a command.
So basically, you can continue using endpoint create, but that does it in the topic-per-event type way and if you want to access the "old commands" you have to do "migrate create" or migrate subscribe etc. which that it would also follow more logically what we have in the transport, and it is better optimized for removal because you can delete an entire command tree when migration is no longer supported

src/CommandLine/Program.cs
```diff
@@ -38,6 +38,44 @@ static int Main(string[] args)
 
             app.HelpOption(inherited: true);
 
+            void SubscribeTopicPerEventType(CommandLineApplication subscribeCommand)
+            {
+                subscribeCommand.Description = "Subscribes an endpoint to an event using topic-per-event topology.";
+                var name = subscribeCommand.Argument("name", "Name of the endpoint (required)").IsRequired();
+                var topicName = subscribeCommand.Argument("topic", "Topic name to subscribe. Unless configured otherwise on the publisher defaults to full name of the event (e.g. MyNamespace.MyMessage) (required)").IsRequired();
```
danielmarbach, line 45: I got a bit confused here because by looking at the description I put myself into the shoes of the user of the command line tool and the wordings says "defaults" which to me indicated the command argument is indeed optional

src/CommandLine/Program.cs
```diff
@@ -38,6 +38,44 @@ static int Main(string[] args)
...
+
+                    Console.WriteLine($"Endpoint '{name.Value}' subscribed to '{topicName.Value}'.");
+                });
+            }
+
+            void UnsubscribeTopicPerEventType(CommandLineApplication unsubscribeCommand)
+            {
+                unsubscribeCommand.Description = "Unsubscribes an endpoint from an event using topic-per-event topology.";
+                var name = unsubscribeCommand.Argument("name", "Name of the endpoint (required)").IsRequired();
+                var topicName = unsubscribeCommand.Argument("topic", "Topic name to subscribe. Unless configured otherwise on the publisher defaults to full name of the event (e.g. MyNamespace.MyMessage) (required)").IsRequired();
```
danielmarbach, line 65: Same comment as above

src/CommandLine/Program.cs
```diff
@@ -49,79 +87,151 @@ static int Main(string[] args)
...
-                    var topicName = unsubscribeCommand.Option("-t|--topic", "Topic name to unsubscribe from (defaults to 'bundle-1')", CommandOptionType.SingleValue);
-                    topicName.DefaultValue = Topic.DefaultTopicName;
-                    var subscriptionName = unsubscribeCommand.Option("-b|--subscription", "Subscription name (defaults to endpoint name) ", CommandOptionType.SingleValue);
-                    var shortenedRuleName = unsubscribeCommand.Option("-r|--rule-name", "Rule name (defaults to event type) ", CommandOptionType.SingleValue);
-
-                    unsubscribeCommand.OnExecuteAsync(async ct =>
+                    endpointCommand.Command("subscribe", subscribeCommand =>
+                        {
+                            subscribeCommand.Description =
+                                "Subscribes an endpoint to an event using legacy forwarding topology.";
```
danielmarbach, line 175: Let's use the term migration here

src/CommandLine/Program.cs
```diff
@@ -49,79 +87,151 @@ static int Main(string[] args)
...
+
+                                Console.WriteLine($"Endpoint '{name.Value}' subscribed to '{eventType.Value}'.");
+                            });
+                        });
+
+                    endpointCommand.Command("unsubscribe", unsubscribeCommand =>
                     {
-                        await CommandRunner.Run(connectionString, fullyQualifiedNamespace, client => Endpoint.Unsubscribe(client, name, topicName, subscriptionName, eventType, shortenedRuleName));
+                        unsubscribeCommand.Description =
+                            "Unsubscribes an endpoint from an event using legacy forwarding topology.";
```
danielmarbach, line 205: Same here

src/CommandLine/Topic.cs
```diff
@@ -6,10 +6,8 @@
 
     static class Topic
     {
-        public static Task Create(ServiceBusAdministrationClient client, CommandOption topicName, CommandOption<int> size, CommandOption partitioning)
+        public static Task Create(ServiceBusAdministrationClient client, string topicNameToUse, CommandOption<int> size, CommandOption partitioning)
```
danielmarbach, line 9: Discussed the inconsistency of the API shapes here.

src/CommandLine/TopicPerEventTopologyEndpoint.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    public static async Task Create(ServiceBusAdministrationClient client, CommandArgument name, CommandOption<int> size, CommandOption partitioning)
+    {
+        try
+        {
+            await Queue.Create(client, name, size, partitioning);
+        }
+        catch (ServiceBusException ex) when (ex.Reason == ServiceBusFailureReason.MessagingEntityAlreadyExists)
+        {
+            Console.WriteLine($"Queue '{name.Value}' already exists, skipping creation");
```
danielmarbach, line 19: Did we use this before? Would it be better to use Console.Error?

danielmarbach (approved):

src/CommandLine/TopicPerEventTopologyEndpoint.cs
SzymonPobiega: I've checked and this is how it worked. Moreover, these are not really error messages because the commands complete successfully so I'll leave it that way.

PR #1173, opened 2025-02-14
Make the link clickable and slightly move the obsoletion

danielmarbach:


src/Transport/EventRouting/MigrationTopology.cs
```diff
@@ -188,7 +188,7 @@ protected override string GetPublishDestinationCore(string eventTypeFullName)
     internal override SubscriptionManager CreateSubscriptionManager(SubscriptionManagerCreationOptions creationOptions) => new MigrationTopologySubscriptionManager(creationOptions, Options);
 
     internal const string ObsoleteMessage =
-        "The migration topology is intended to be used during a transitional period, facilitating the migration from the single-topic topology to the topic-per-event topology. The migration topology will eventually be phased out over subsequent releases. Should you phase challenges during migration, please reach out to https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/1170.";
+        "The migration topology is intended to be used during a transitional period, facilitating the migration from the single-topic topology to the topic-per-event topology. The migration topology will eventually be phased out over subsequent releases. Should you phase challenges during migration, please reach out to |https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/1170|.";
```
ramonsmits, line 191: That `.` at the end, is that needed?

TIL: Using | to get links in obsolete messages

ramonsmits (approved):

ramonsmits:
> how we have set the version means it would no longer work in the LTS version that comes out. Maybe that is a bit too tight isn't it?

Yeah, I think it should be supported till the next LTS. As in, users migrating from LTS to LTS.

src/Transport/EventRouting/MigrationTopology.cs
danielmarbach: The . makes it a real sentence, so I'd prefer to keep it.

ramonsmits: I get it. the || are the equivalent of markdown its <> syntax

SzymonPobiega (approved):

PR #1205, opened 2025-05-28
Use `IsDebugEnabled` to reduce costs of debug log events

ramonsmits:
ILogger.Debug() calls that use string interpolation or invoking methods to get log arguments wrapped in `IsDebugEnabled`

The actual improvement would be to remove all string interpolation and use log template via the `.****Format` methods but that is a lot of work.


danielmarbach (approved):
Some of the flag checks are not really required because the methods are not called on the hot path but I'm ok to lean on the side of consistency

PR #1208, opened 2025-06-12
Fix transport encoding header obsoletion

danielmarbach:
Outcome of a discussion with @bording as a proposal

As part of https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1072 the ability to disable sending the transport encoding headers was introduced and marked as 4.3.0 but never actually released. When @SzymonPobiega and I started working on the changes for v5 we followed through the obsoletion pattern of opt-out, opt-in and then remove for something that was never actually released.

Master was in a weird in-between state that made us probably confused and quite likely we have not spend enough time to connect all the dots again. Now we have v5 shipped containing obsoletes that make little sense. This PR fixes these obsoletes under the following assumption:

- Given that opt-out has never been shipped and v5 already opts-out by default and has the ability to opt-in again we partially follow our regular pattern (there is not much to undo that without unnecessary extra work). We will update the upgrade guide of v5 to make it clear that you can get compatibility with legacy versions by opt-in for another major but this option will go out of support
- The obsoletion messages on the opt-in part need to properly reflect that the API will get out of support in the next major

This should be backported to v4.2 branches


danielmarbach:
> Wait, so this change was never actually released?

Nope. It wasn't followed through. So milestone 4.3 is unreleased

tmasternak (approved):

bording (approved):

Issue #1141, opened 2025-01-27
When connecting to ASB fails, log additional error information to help diagnose and resolve the issue


ramonsmits:
#### Is your feature related to a problem? Please describe.

A customer was having connectivity issues with ServiceControl because the AMQP ports for that server were blocked. Outgoing ports 80 and 443 were allowed.

#### Describe the requested feature

While this can be toggled via the `UseWebSockets` connectionstring argument it would make sense to for example have a flow like:

1. Connect to AMQP
2. If fails, connect to Websockets and log an WARN
3. If fails, raise FATAL (not error)
        
#### Describe alternatives you've considered

Alternatively the exception could state what it can't connect to (AMQP or Websockets or even list ports) and report:

> Can't connected using AMQP protocol, try websockets or ensure AMQP ports 5671 and 5672  are reachable 
 to host xxxxxxxxx

## Related

- https://github.com/Particular/ServiceControl/issues/4751

abparticular:
Will be shipped with release 5.1

PR #1235, opened 2025-08-28
Improve exception messages for connectivity exceptions

abparticular:
Addresses:

- #1141 

In order to provide additional context to connectivity exceptions to ASB, the `MessageDispatcher` class needs access to the `ServiceBusClient` instance that was used to connect, so it can determine whether TCP or WebSockets was used.  For this reason, the default `ServiceBusClient` instance is provided to the `MessageDispatcher` class instead of `MessageSenderRegistry`, and it is the `MessageDispatcher` that coalesces any null client parameter to the default client.

### DIFF
--- a/src/Tests/Sending/MessageDispatcherTests.cs
+++ b/src/Tests/Sending/MessageDispatcherTests.cs
@@ -17,7 +17,7 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Sending
         {
             var client = new FakeServiceBusClient();
 
-            var dispatcher = new MessageDispatcher(new MessageSenderRegistry(client), TopicTopology.FromOptions(new TopologyOptions()));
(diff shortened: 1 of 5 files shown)


PhilBastian (approved):

ramonsmits (approved):
Nice improvement!

PhilBastian:
from a potentially anonymous contributor:
> I was wondering if, for that improvement, it is really worth decorating the actual exception of the SDK and forcing the user to check the inner exception
Technically we could achieve the same by adding a proper log error in that catch and throw the original, which would also keep the client behavior upwards the same since we are keeping the expectations that we throw "underlying SDK exceptions"

ramonsmits:
Liking that reasoning a lot. Not wrapping the exception has benefits especially because a too generic `Exception` is used. Logging an error and rethrowing to me feels like the perfect middle ground.

PR #1241, opened 2025-09-09
Spike for creating the manifest for ASB

PhilBastian:
related to https://github.com/Particular/NServiceBus/pull/7406

PR #1248, opened 2025-09-23
add manifest information to diagnostics

PhilBastian:
alternative to https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1241

related to https://github.com/Particular/NServiceBus/pull/7416


src/Transport/EventRouting/TopicPerEventTopologySubscriptionManager.cs
```diff
@@ -14,27 +14,47 @@ namespace NServiceBus.Transport.AzureServiceBus;
...
+                .GetValueOrDefault(eventTypeFullName, [eventTypeFullName])
+                .Select(topicName => new { Topic = topicName.ToLower(), MessageType = eventTypeFullName }))
+            .GroupBy(topicAndMessageType => topicAndMessageType.Topic)
+            .Select(group => new
+            {
+                TopicName = group.Key,
+                MessageTypes = group.Select(topicAndMessageType => topicAndMessageType.MessageType).ToArray()
+            })
+            .ToArray();
+        startupDiagnostic.Add("Manifest-Subscriptions", subscriptions);
```
jpalac, line 47: Is it worth pulling this out into it's own method/extension since we are repeating the same code in each SubscriptionManager?

PhilBastian: I considered it, but MigrationTopology is getting deleted in the next major

PR #1309, opened 2026-01-08
Add support for hierarchical entity names (namespace prefixing)

### LINKED ISSUE #1013: Support for hierarchical entities
#### Is your feature related to a problem? Please describe.

Azure Service Bus allows entities to be organized in hierarchies on the same namespace. This can be handy to group for example tenant related queues and topics together. 

So, for example, you can have a hierarchy like this

tenanta/endpoint1
tenanta/endpoint2
tenantb/endpoint1
tenantb/endpoint2

image

To achieve that, there are a number of "quirks" required on an endpoint configuration level.

The trick is to override the local queue address to contain the hierarchy and make sure the routing table contains the hierarchy too. For separate audit and error queues, you also need to make sure the audit and error queues are configured accordingly. 

```csharp
        var tenantIdentifier = Environment.GetEnvironmentVariable("TenantIdentifier");

        var endpointConfiguration = new EndpointConfiguration($"Endpoint1");
        // Override the local address
        endpointConfiguration.OverrideLocalAddress($"{tenantIdentifier}/Endpoint1");
        endpointConfiguration.EnableInstallers();

        // Override the diagnostic path to have unique paths for endpoint for demo purposes
        endpointConfiguration.SetDiagnosticsPath($".diagnostics/{tenantIdentifier}");

        // configure tenant specific error queue
        endpointConfiguration.SendFailedMessagesTo($"{tenantIdentifier}/error");
        // configure tenant specific audit queue
        endpointConfiguration.AuditProcessedMessagesTo($"{tenantIdentifier}/audit");

        // switch the topology to have a bundle per tenant
        // make sure the subscription doesn't contain illegal characters in the path
        var transport = new AzureServiceBusTransport(connectionString)
        {
            Topology = TopicTopology.Single($"{tenantIdentifier}/bundle1"),
            SubscriptionNamingConvention = subscription => subscription.Replace($"{tenantIdentifier}/", string.Empty)
        };
        var routing = endpointConfiguration.UseTransport(transport);
        // Carefully override each command routing to go to the right destination
        routing.RouteToEndpoint(typeof(Command1), $"{tenantIdentifier}/Endpoint2");
```

There is currently one downside to this approach. You have to be careful to configure the tenant identifier on all the sends; otherwise you might send to some "central" queue.

SQS as the concept of [QueueNamePrefix](https://docs.particular.net/transports/sqs/configuration-options#queue-name-prefix) that is applied on a transport level, which makes "isolating" the same endpoints on the same SQS account a breeze and doesn't require changing the routing tables, error and audit queues.

#### Describe the requested feature

Add a possibility to leverage hierarchies to facilitate grouping of the same endpoints within a hierarchy.
        
#### Describe alternatives you've considered

None

### Additional Context

Sample implementation

[TenantsWithAsbUsingHierarchy.zip](https://github.com/user-attachments/files/16426724/TenantsWithAsbUsingHierarchy.zip)

DharmishthaVPatel:
Closes #1013

This allows entities in Azure Service Bus to be organized into hierarchies on the same namespace using a `HierarchyNamespaceOptions` property:

```csharp
transport.HierarchyNamespaceOptions = new HierarchyNamespaceOptions
{
    HierarchyNamespace = "my-hierarchy"
};
```

Setting this prefixes all message destinations with the  `HierarchyNamespace` followed by a `/`:

```
my-hierarchy/destination
```

Because the prefix is applied to all messages, topology mappings do not need to specify the hierarchy namespace to utilize this feature.  This also means that topology mappings are unable to "escape" the hierarchy if it is set.  If a topology mapping attempts to set the same hierarchy prefix, it will be ignored, but if it attempts to set a different one, it will still be prepended with the HierarchyNamespace. For example, if a topology mapping specifies a destination of `topologyns/destination`, the destination using the above options would become:

```
my-hierarchy/topologyns/destination
```

For scenarios where this blanket hierarchy approach is not desired, there are two options:
- Omit the HierarchyNamespace property and configure topology mappings for destinations where a hierarchy is desired. This is better when most messages will not be within the hierarchy.
- Set the HierarchyNamespaceOptions property and configure message types that should be excluded from the hierarchy:
  ```csharp
  // A single message type
  transport.HierarchyNamespaceOptions.ExcludeMessageType<MyMessage>();
  
  // Multiple types at once using interfaces or an inherited base class
  transport.HierarchyNamespaceOptions.ExcludeMessageType<ISomeInterface>();
  ``` 
  This is better when most messages need to be in the hierarchy.

### DIFF
--- a/src/Transport/AzureServiceBusTransport.cs
+++ b/src/Transport/AzureServiceBusTransport.cs
@@ -219,7 +219,6 @@ public partial class AzureServiceBusTransport : TransportDefinition
     /// <summary>
     /// The maximum size used when creating queues and topics in GB.
     /// </summary>
-    /// 
     public int EntityMaximumSize
     {
         get;


SeanFeldman:
Related to #1013

ngallegos:
@SeanFeldman , we'd welcome any thoughts or insights you have on this as well, based on [your interest in the issue](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/1013#issuecomment-3487848049).

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -82,7 +84,7 @@ public override async Task<TransportInfrastructure> Initialize(HostSettings host
             {
                 TransportType = transportType,
                 EnableCrossEntityTransactions = enableCrossEntityTransactions,
-                Identifier = $"Client-{receiver.Id}-{receiver.ReceiveAddress}-{Guid.NewGuid()}",
+                Identifier = $"Client-{HierarchyNamespaceClientIdentifier}{receiver.Id}-{receiver.ReceiveAddress}-{Guid.NewGuid()}"
```
danielmarbach: Won't this look weird with slashes in the name? It might be fine; I just wanted to mention it to check whether you have thought about that angle and considered replacing the slashes in the hierarchy with "-" or something else to make it a more human-readable identifier (after all, this will appear in the event logs).

danielmarbach:
An additional angle that I see missing is considerations around the command line tool. Should the prefix be something the tool accepts as an additional parameter or is it something the user has to specify as part of the queue / topic?

src/Transport/AzureServiceBusTransport.cs
ngallegos: Good catch, we hadn't considered HierarchyNamespaces with additional `/` separators. This has been addressed.

src/CommandLineTests/CommandLineTests.cs
```diff
@@ -277,6 +488,28 @@ public async Task Unsubscribe_migration_endpoint_supports_hierarchy()
...
+        public async Task Unsubscribe_hierarchy_namespace_migration_endpoint_supports_hierarchy()
+        {
+            var queueName = QueueName.ToHierarchyNamespaceAwareDestination(HierarchyNamespace);
+            var topicName = TopicName.ToHierarchyNamespaceAwareDestination(HierarchyNamespace);
+            var hierarchyTopicName = HierarchyTopicName.ToHierarchyNamespaceAwareDestination(HierarchyNamespace);
+            await DeleteQueue(queueName);
+            await DeleteTopic(topicName);
+            await DeleteTopic(hierarchyTopicName);
+
+            await Execute($"migration endpoint create {EndpointName} --topic-to-publish-to {TopicName} --topic-to-subscribe-on {HierarchyTopicName} --hierarchy-namespace {HierarchyNamespace}");
```
danielmarbach: Commenting only on this test here but the question quite likely also applies to others. What effect does the hierarchy-namespace option supposed to have here? If I understand we are currently saying that it should "shift" the migration endpoint consistently down into the hierarchy. I think that might work. Although you could also argue that since a migration endpoint might need to indicate exactly where to subscribe to the topic-to-publish-to and subscribe-on need to be fully qualified topics, including the hierarchy like your test verifies. I think that is consistent. Will you describe this in the docs so that is is clear? Sorry, just trying to understand the scenario and making sure I understand it correctly.

src/Transport/EventRouting/DestinationManager.cs
```diff
@@ -0,0 +1,48 @@
...
+                    return destination;
+                }
+
+                if (messageTypeFullNames is { Length: > 0 } && hierarchyNamespaceOptions.MessageTypeFullNamesToExclude.Any(messageTypeFullNames.Contains))
+                {
+                    return destination;
+                }
+
+                var hierarchyNamespace = hierarchyNamespaceOptions.HierarchyNamespace;
+                var hierarchyNamespaceSpan = string.Concat(hierarchyNamespace, '/').AsSpan();
```
danielmarbach: This concat seems unnecessary and always be the same. It is possible to hoist ReadOnlySpans in statics and use it elsewhere. I can show that in some code snippet

src/AcceptanceTests/Sending/When_using_hierarchy_namespace.cs
```diff
@@ -0,0 +1,123 @@
...
+                .WithEndpoint<ExternalReceiver>()
+                .WithEndpoint<HierarchyReceiver>(endpoint =>
+                {
+                    endpoint.CustomConfig(cfg =>
+                    {
+                        var transport = cfg.ConfigureTransport<AzureServiceBusTransport>();
+                        transport.HierarchyNamespaceOptions = new HierarchyNamespaceOptions { HierarchyNamespace = "my-hierarchy" };
+                    });
+                })
+                .Done(c => c.ExternalMessageReceived)
```
danielmarbach: Same here

src/AcceptanceTests/Sending/When_using_hierarchy_namespace.cs
```diff
@@ -0,0 +1,123 @@
...
+                })
+                .WithEndpoint<HierarchyReceiver>(endpoint =>
+                {
+                    endpoint.CustomConfig(cfg =>
+                    {
+                        var transport = cfg.ConfigureTransport<AzureServiceBusTransport>();
+                        transport.HierarchyNamespaceOptions = new HierarchyNamespaceOptions { HierarchyNamespace = "my-hierarchy" };
+                    });
+                })
+                .Done(c => c.HierarchyMessageReceived)
```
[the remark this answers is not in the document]
danielmarbach: Also wouldn't it be better to split those tests because that would allow you to call MarkAsFailed when the external handler receives it when the hierarchy should not be escaped?

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -118,8 +119,16 @@ await administrationClient.AssertNamespaceManageRightsAvailable(cancellationToke
 
             var allQueues = infrastructure.Receivers
                 .Select(r => r.Value.ReceiveAddress)
-                .Concat(sendingAddresses)
+                .Concat(sendingAddresses.Select(s => DestinationManager.GetDestination(s)))
                 .ToArray();
+            // Validate the actual names that will be created (including hierarchy namespace prefix)
+            var validationResult = EntityValidator.ValidateQueues(allQueues, nameof(TopologyOptions.QueueNameToSubscriptionNameMap));
```
danielmarbach: Something feels off. The transport does validate the event topology early on a few lines above. That  always happens independent of whether the setup infrastructure is on. I think the idea to validate the queue is sound and great but should we then consolidate the validation into a sensible place to happen even when the setup infrastructure flag is no set?

Issue #1104, opened 2024-11-28
Add support for the Azure Service Bus emulator


boblangley:
#### Describe the requested feature

Microsoft [released a local emulator for Azure Service Bus](https://techcommunity.microsoft.com/blog/messagingonazureblog/introducing-local-emulator-for-azure-service-bus/4304457).

Currently the emulator [lacks some features](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator#known-limitation) that make using the emulator in a development scenario with NServiceBus difficult or impossible:

- The emulator does not support any [admin features from the SDK](https://github.com/Azure/azure-service-bus-emulator-installer/issues/17). Entities must be setup beforehand [using a configuration file](https://github.com/Azure/azure-service-bus-emulator-installer/blob/main/ServiceBus-Emulator/Config/Config.json) and [NServiceBus installers cannot be run without errors](https://docs.particular.net/nservicebus/operations/installers).
- [The emulator does not support SQL filters](https://github.com/Azure/azure-service-bus-emulator-installer/issues/17), breaking the NServiceBus pub/sub implementation.

This issue can help track when Particular Software becomes aware that these features have been added to the emulator and when this transport is then compatible with it.

#### Describe alternatives you've considered

The current alternative is to continue using a full Azure namespace for development. Many users use a Standard tier namespace for development, often one per developer due to it's relatively low cost, and move to a Premium tier namespace for production.

SeanFeldman:
Original thread: [forum](https://discuss.particular.net/t/error-using-azureservicebustransport-with-microsoft-azure-service-bus-emulator/4175).

dlbr-pci:
A recent release of the emulator, [version 1.1.2](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-emulator-whats-new#latest-version-112
) states: 
> SQL Actions and Filters are now compatible with Service bus emulator.

Does this make the feature now possible (albeit without _installers_ support)?

SeanFeldman:
> Does this make the feature now possible (albeit without _installers_ support)?

In theory, yes.

danielmarbach:
[...]

PR #1331, opened 2026-02-20
Add proper support for Azure Service Bus Emulator for development and testing

yngvebn:
Add support for configuring MaxDeliveryCount and AdministrationClientOptions
Related issue #1104

This has been tested and verified on services running on version 5.1.2 of NServiceBus.Transport.AzureServiceBus (as we're not yet on .net 10, we cannot update to 6+ yet)

Questions for the maintainers: Do you want me to also implement the changes as a separate pullrequest on the latest code?

---

Since Azure Service Bus emulator now exposes the management API, the emulator should now be possible to fully replace learning transport or real azure service bus for testing and development purposes. (Jeremy Miller writes about it here: https://jeremydmiller.com/2026/02/19/using-the-azure-service-bus-emulator-for-local-wolverine-development/)

The emulator exposes the management api on a separate endpoint, so in order to use it it requires the endpoint adress to be explicitly configurable.

Additionally, the Azure Service Bus emulator imposes constraints not present in the real service: it caps MaxDeliveryCount at 10.

- Add `MaxDeliveryCount` property to `AzureServiceBusTransport` (defaults to `int.MaxValue` for backwards compatibility) and apply it consistently when creating queues and subscriptions across all topology creators and subscription managers.

- Add `AdministrationClientOptions` property to allow consumers to customise the `ServiceBusAdministrationClient`, e.g. to provide a custom HTTP transport redirecting requests to a non-TLS emulator endpoint or an Endpoint connectionstring as exposed by Azure Service Bus Emulator.

Fixes startup failures when using the Azure Service Bus emulator with `SetupInfrastructure = true`.

### DIFF
--- a/src/Tests/Administration/QueueCreatorTests.cs
+++ b/src/Tests/Administration/QueueCreatorTests.cs
@@ -27,6 +27,24 @@ public class QueueCreatorTests
     }
 
     [Test]
+    public async Task Should_use_configured_MaxDeliveryCount()
+    {
+        var transport = new AzureServiceBusTransport("connectionString", TopicTopology.Default)
+        {
+            MaxDeliveryCount = 10
+        };
+
+        var recordingClient = new RecordingServiceBusAdministrationClient();
+        var creator = new QueueCreator(transport);
+
+        await creator.Create(recordingClient, ["test-queue"], "test-queue");
+
+        var output = recordingClient.ToString();
+
+        Approver.Verify(output);
+    }
+
+    [Test]
     public async Task Should_not_set_AutoDeleteOnIdle_when_null()
     {
         var transport = new AzureServiceBusTransport("connectionString", TopicTopology.Default)
--- a/src/Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
+++ b/src/Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
@@ -16,10 +16,12 @@ namespace NServiceBus
             "ersion 6.0.0.", true)]
         public AzureServiceBusTransport(string fullyQualifiedNamespace, Azure.Core.TokenCredential tokenCredential) { }
         public AzureServiceBusTransport(string fullyQualifiedNamespace, Azure.Core.TokenCredential tokenCredential, NServiceBus.TopicTopology topology) { }
+        public Azure.Messaging.ServiceBus.Administration.ServiceBusAdministrationClientOptions? AdministrationClientOptions { get; set; }
         public System.TimeSpan? AutoDeleteOnIdle { get; set; }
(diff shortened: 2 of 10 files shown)


yngvebn:
> @yngvebn Thank you for taking the time to contribute to the codebase!
> 
> It would be best to target the pull request against master. We can then decide on the backporting options as a separate decision.
> 
> I have not yet looked into all the details of the PR yet. I have two things, though, that I would like to discuss:
> 
> * Generally we don't want to expose settings that could impose problems with the higher level APIs. So for example, exposing the max delivery count is problematic since that could cause ripple effects with the recoverability policy.
> * Given the emulator uses a well defined connection string "UseDevelopmentEmulator=true" have you considered special casing the transport internals against that? It seems the SDK parser already accounts for the special connection string and it would allow us to discover the emulator. Currently the [property is still internal](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusConnectionStringProperties.cs#L98), but we can easily access that with UnsafeAccessor and/or contribute to the SDK to make this public. I also think both the service bus client and admin client automatically adjust to the emulator, see for example https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Administration/ServiceBusAdministrationClient.cs#L109
> 
> All this makes me think that we don't need to change the public API surface of the transport. We can switch based on the connection string to support the emulator and override internally the delivery count manually.
> 
> Please let me know if you'd like to continue driving this change or whether you'd like us to take over things at some point.
> 
> Thoughts?

Ah , yes this makes sense. I totally missed the stuff from the ServiceBusAdministrationClient source. I guess auto setting the MaxDeliveryCount based on the connection-string containing UseDevelopmentEmulator=true flag would be fine, but that's more up to you guys really. This would introduce hidden magic that's invisible for the user, but I recon that would be fine in the case of emulator/dev/testing? Perhaps a log-message telling the user that MaxDeliveryCount is capped at 10 due to emulator or something?

Instead of UnsafeAccessor or contributing to the SDK, a string check would probably also suffice?

I don't mind driving the change forward. It kind of depends on your capacity, and if it makes any difference release/time wise. We'd really like this functionality, cause currently we need to spin up a complete Azure Service Bus resource group in Azure in order to run dev-containers. That's a bit of a hassle :D

danielmarbach:
@yngvebn Feel free to draw inspiration from https://github.com/Particular/setup-azureservicebus-action or use it. We are not officially supporting this for our customers, but every change we do to these repositories is well versioned. This is how we spin up a dynamic ASB namespace for every build and then remove it again

PR #1342, opened 2026-03-09
Add performance test infrastructure for Azure Service Bus

dvdstelt:
Based on the new NuGet Sources package created here: https://github.com/Particular/NServiceBus.PerformanceTests


danielmarbach (commented):
I think it would benefit from some small instructions in the readme on when to run those and what to expect

dvdstelt:
I will update this

PR #1368, opened 2026-04-03
Fallback to inferred and deterministic message ID when Service Bus message ID is missing

andreasohlund:
This aligns the behavior with the new Azure function package and makes the transport more resillient


src/TransportTests/When_processing_a_message_sent_without_explicit_message_id.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+using Azure.Messaging.ServiceBus;
+using NServiceBus.TransportTests;
+using NUnit.Framework;
+
+[TestFixture]
+public class When_processing_a_message_sent_without_explicit_message_id : NServiceBusTransportTest
```
andreasohlund, line 11: @danielmarbach turns out that it wasn't the portal that defaulted the message ID, but rather the client (or broker). This means that we can never get a null or empty message Id but I still think it makes sense to handle this as long as the API exposes the MessageId as nullable?

danielmarbach (approved):

DavidBoike (approved):

PR #1371, opened 2026-04-07
Fix subscription name overrides being ignored when using a hierarchy namespace

### LINKED ISSUE #1369: Subscription name override not applied when using hierarchy namespaces
#### Description

When using `TopicPerEventTopologySubscriptionManager` together with hierarchy namespaces, subscription name overrides may not be applied correctly.

The issue occurs because the override lookup is performed using the namespaced subscribing queue name (for example `my-hierarchy/SubscribingQueue`) while overrides are typically registered using the non-namespaced endpoint name (for example `SubscribingQueue`).

As a result, the lookup fails and the transport falls back to the default derived subscription name instead of using the configured override.

#### Expected behavior

Subscription name overrides should be correctly applied regardless of whether hierarchy namespaces are enabled.

If an override is configured for a queue name like `SubscribingQueue`, it should also be honored when the runtime subscribing queue name is namespaced (for example `my-hierarchy/SubscribingQueue`).

#### Actual behavior

When hierarchy namespaces are enabled, the override lookup uses the namespaced queue name, which does not match the keys in `QueueNameToSubscriptionNameMap`.

This causes the override to be missed, and the transport falls back to the default subscription name instead of the intended overridden value.

#### Versions

- Affects: `NServiceBus.Transport.AzureServiceBus` versions supporting hierarchical entities

### Steps to reproduce

```csharp
    [Test]
    public async Task Should_apply_subscription_name_override_for_non_namespaced_queue_when_hierarchy_enabled()
    {
        var hierarchyOptions = new HierarchyNamespaceOptions { HierarchyNamespace = "my-hierarchy" };
        var topologyOptions = new TopologyOptions
        {
            HierarchyNamespaceOptions = hierarchyOptions,
            SubscribedEventToTopicsMap =
            {
                { typeof(MyEvent1).FullName, ["MyTopic"] }
            },
            QueueNameToSubscriptionNameMap = { { "SubscribingQueue", "MySubscriptionName" } },
        };

        var builder = new StringBuilder();
        var client = new RecordingServiceBusClient(builder);
        var administrationClient = new RecordingServiceBusAdministrationClient(builder);

        var subscriptionManager = new TopicPerEventTopologySubscriptionManager(new SubscriptionManagerCreationOptions
        {
            SubscribingQueueName = "my-hierarchy/SubscribingQueue",
            Client = client,
            AdministrationClient = administrationClient
        }, topologyOptions, new StartupDiagnosticEntries());

        await subscriptionManager.SubscribeAll([new MessageMetadata(typeof(MyEvent1))], new ContextBag());

        Assert.That(builder.ToString(), Does.Contain("\"SubscriptionName\": \"MySubscriptionName\""));
    }
```

### Relevant log output

```shell

```

### Symptoms

Subscription name overrides are not applied when using hierarchical namespaces.

### Who's affected

Users of [hierarchical namespaces](https://docs.particular.net/transports/azure-service-bus/configuration#entity-creation-hierarchy-namespace) and [subscription name overrides](https://docs.particular.net/transports/azure-service-bus/configuration#entity-creation-topology-mapping) that omit the hierarchy in the queue names keys.

### Fix Available

A fix is available in [NServiceBus.Transport.AzureServiceBus 6.2.2](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/releases/tag/6.2.2)

ngallegos:
Fixes #1369

The current topology subscription manager implementations assume that subscription name overrides are aware of hierarchy namespaces.  This goes against the idea behind the hierarchy namespace implementation, which is to have a single configuration instead of propagating the hierarchy everywhere.

This aligns the topology subscription managers with the rest of the hierarchy namespace implementation, specifically, allowing users to configure overrides within a hierarchy as expected:

```csharp
var hierarchyOptions = new HierarchyNamespaceOptions { HierarchyNamespace = "my-hierarchy" };
var topologyOptions = new TopologyOptions
{
    HierarchyNamespaceOptions = hierarchyOptions,
    SubscribedEventToTopicsMap =
    {
        { typeof(MyEvent1).FullName, ["MyTopic"] }
    },
    QueueNameToSubscriptionNameMap = { { "SubscribingQueue", "MySubscriptionName" } },
};
```

rather than having to remember to add the hierarchy to the overrides (QueueNameToSubscriptionNameMap):

```csharp
var hierarchyOptions = new HierarchyNamespaceOptions { HierarchyNamespace = "my-hierarchy" };
var topologyOptions = new TopologyOptions
{
    HierarchyNamespaceOptions = hierarchyOptions,
    SubscribedEventToTopicsMap =
    {
        { typeof(MyEvent1).FullName, ["MyTopic"] }
    },
    QueueNameToSubscriptionNameMap = { { "my-hierarchy/SubscribingQueue", "MySubscriptionName" } },
};
```


src/Tests/EventRouting/TopicPerEventSubscriptionManagerTests.cs
danielmarbach: Looking at the test name, what is this actually testing? Or in other words, why would we ever want to strip things from subscription names? I think the idea was, probably, that we want to test that the hierarchy doesn't influence the subscription name, but that seems to be already covered by the other two tests we added, or am I seeing this from the wrong angle?

src/Tests/EventRouting/MigrationTopologySubscriptionManagerTests.cs
danielmarbach: Looking at the test name, what is this actually testing? Or in other words, why would we ever want to strip things from subscription names? I think the idea was, probably, that we want to test that the hierarchy doesn't influence the subscription name, but that seems to be already covered by the other two tests we added, or am I seeing this from the wrong angle?

src/Tests/EventRouting/TopicPerEventSubscriptionManagerTests.cs
ngallegos: I agree, really, what we are testing is that the hierarchy doesn't influence subscription names, so this test could be named better.

I do think it's worth keeping because the new tests are testing that the hierarchy doesn't affect the subscription name overrides. This test used to do that, but now it tests that the default subscription names are not affected.

src/Tests/EventRouting/TopicPerEventSubscriptionManagerTests.cs
```diff
@@ -76,7 +76,6 @@ public async Task Should_strip_hierarchy_namespace_from_subscription_names()
             {
                 { typeof(MyEvent1).FullName, [ destinationManager.GetDestination("MyTopic1"), destinationManager.GetDestination("MyTopic2")] }
             },
-            //QueueNameToSubscriptionNameMap = { { "SubscribingQueue", "MySubscriptionName" } },
```
danielmarbach: Would it still be helpful to keep the map so that the approval file contains `MySubscriptionName` instead of `SubscribingQueue`? This is what the original approval file contained

danielmarbach: Same for the other test in the Migration part

danielmarbach (approved):
My two last comments are just thoughts and not worth "blocking" an approval

src/Tests/EventRouting/TopicPerEventSubscriptionManagerTests.cs
danielmarbach: I also think you need to rename the approval files accordingly

ngallegos: I was keeping the map out of these tests so they can cover the default naming scenario. If we add it back, then  it would be testing the same logic as the new tests that have the map.

PR #1373, opened 2026-04-07
Add support for native dead-lettering and DLQ auto-forwarding

andreasohlund:
This change gives users the ability to fully utilize the dead lettering feature of Azure Service Bus by:

- Providing API's to request failing messages to be dead-lettered
- Optionally provision queues with dead letter forwarding

## API to explicitly request dead lettering

```c#
endpointConfiguration.Recoverability()
   .CustomPolicy((config, errorContext) =>
   {
       if (errorContext.Exception is MyException ex)
       {
           return RecoverabilityAction.DeadLetter(
               deadLetterReason: "Some reason",
               deadLetterErrorDescription: ex.Message,
               propertiesToModify: new Dictionary<string, object>
               {
                   ["MyCustomPropery"] = "Some value"
               });
       }

       return DefaultRecoverabilityPolicy.Invoke(config, errorContext);
   });
```
If you just want to dead-letter with the same faults metadata as NSerivceBus would add:
```c#
endpointConfiguration.Recoverability()
   .CustomPolicy((config, errorContext) =>
   {
       if (errorContext.Exception is PoisonMessageException)
       {
           return RecoverabilityAction.DeadLetter();
       }

       return DefaultRecoverabilityPolicy.Invoke(config, errorContext);
   });
```

## API to use native DLQ for failures

To route failures via the native dead-letter queue, which is more performant and cost effective use:

```
var recoverability = endpointConfiguration.Recoverability();
recoverability.MoveErrorsToAzureServiceBusDeadLetterQueue();
```

## API to enable dead letter forwarding when queues are created

Via the endpoint config:

```c#
var endpointConfiguration = new EndpointConfiguration("Sales");
 endpointConfiguration.SendFailedMessagesTo("error");
 
 var transport = new AzureServiceBusTransport(
     "<connection string>",
     TopicTopology.Default)
 {
     AutoForwardDeadLetteredMessagesToErrorQueue = true
 };
 
 endpointConfiguration.UseTransport(transport);
```

Using the command line tool:
```con
# Queue create

asb-transport queue create sales \
 --connection "<connection string>" \
 --forward-dlq-to error

# Endpoint create

asb-transport endpoint create sales \
 --connection "<connection string>" \
 --forward-dlq-to error

# Migration endpoint create

asb-transport migration endpoint create sales \
 --connection "<connection string>" \
 --forward-dlq-to error

```

## Fault headers mapping

When processing a dead-lettered message, the following headers are set unless already present:

 - `ServiceBusReceivedMessage.DeadLetterSource` => `NServiceBus.FailedQ`
 - `ServiceBusReceivedMessage.DeadLetterReason` => `NServiceBus.ExceptionInfo.Message`
 -  `ServiceBusReceivedMessage.DeadLetterErrorDescription` => `NServiceBus.ExceptionInfo.StackTrace`

This allows tools like ServicePulse to better visualize failures.

## Migration

The dead letter forwarding setting is not visible in the portal. Use the following command to view the current status:

```
az servicebus queue list \
  --resource-group MyResourceGroup \
  --namespace-name MyServiceBusNamespace \
--query "[].{name:name,dlqforwarding:forwardDeadLetteredMessagesTo}" \ 
--output table
```
Update endpoint queues using:

```
az servicebus queue update \
  --resource-group MyResourceGroup \
  --namespace-name MyServiceBusNamespace \
  --name MyEndpoint \
  --forward-dead-lettered-messages-to MyErrorQueue
```

### Only allow dead-lettering during recoverability

To be consistent with existing recoverability behavior, only messages that fail processing can be dead-lettered. 

### Make auto forwarding opt-in

To avoid a change in behavior, dead letter forwarding is opt-in; this will be changed in the next major.

### Truncation of dead letter reason and description

To reduce the risk of message oversize, the dead-letter reason and description are each truncated to 1024.


src/Tests/.editorconfig
```diff
@@ -5,3 +5,7 @@ dotnet_diagnostic.CA2007.severity = none
 
 # Justification: Tests don't support cancellation and don't need to forward IMessageHandlerContext.CancellationToken
 dotnet_diagnostic.NSB0002.severity = suggestion
+
+
```
andreasohlund:
```suggestion
```

ramonsmits:
@andreasohlund isn't it possible to add a helper like:

```c#
transport.UseDeadLetterQueues();
```

This is a lot of ceremony.

andreasohlund:
>This is a lot of ceremony.

Clarified with @ramonsmits and https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1377 explores this option

src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -35,6 +37,41 @@ static class MessageExtensions
...
         }
 
+        if (!string.IsNullOrWhiteSpace(message.DeadLetterSource) && !headers.ContainsKey(FaultsHeaderKeys.FailedQ))
+        {
+            headers[FaultsHeaderKeys.FailedQ] = message.DeadLetterSource;
+        }
+
+        if (!string.IsNullOrWhiteSpace(message.DeadLetterReason))
+        {
+            var parts = message.DeadLetterReason.Split(DeadLetterRequest.Separator);
```
DavidBoike: Considered asking if we should be doing IndexOf and Span related things, but since this only applies messages that are going to be dead-lettered, that would make this outside the critical path, right?

DavidBoike (approved):

DavidBoike:
Curious what you decided about something to warn the user that dead letter forwarding would turn on by default in the next major? Or should it be a 3-way switch Enable/Disable/Default so that you could, in this version, say "in the next version I do NOT want that to happen"?

src/Transport/Receiving/MessageExtensions.cs
andreasohlund: See https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1377/

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -176,6 +163,52 @@ void ApplyWebProxyIfNeeded(ServiceBusClientOptions options)
...
+        }
+
+        // make sure that queues without forwarding are created first.
+        return [.. queues.Values.OrderBy(static queue => queue.ForwardDeadLetteredMessagesTo is null ? 0 : 1)];
+
+        CreateQueueOptions BuildDefaultCreateQueueOptions(string queueName) => new(queueName)
+        {
+            EnableBatchedOperations = true,
+            LockDuration = TimeSpan.FromMinutes(5),
+            MaxDeliveryCount = MaxDeliveryCount,
```
andreasohlund, line 218: @DavidBoike @danielmarbach I had a chat with @ramonsmits, and he suggested that with DLQ forwarding enabled, we can assume that users are monitoring their DLQs (which we think was the reason we went with Int.MaxValue) and therefore go with a less insane default, let's say 100?

Technically we only need it to be `> Max(DelayedRetries)`

src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -35,6 +37,41 @@ static class MessageExtensions
             headers[Headers.ContentType] = message.ContentType;
         }
 
+        if (!string.IsNullOrWhiteSpace(message.DeadLetterSource) && !headers.ContainsKey(FaultsHeaderKeys.FailedQ))
+        {
+            headers[FaultsHeaderKeys.FailedQ] = message.DeadLetterSource;
```
andreasohlund, line 41: Until SC is updated to the new version, messages dlq'd by the service bus it self will result in:


```
026-04-11 08:58:18.9321|00:08:30.458|3|Warn|ServiceControl.Operations.ErrorIngestor|Processing of message '2b925812-3cba-4a95-b629-3688882ea9f3' failed|System.Exception: Missing 'NServiceBus.FailedQ' header. Message is poison message or incorrectly send to (error) queue.

   at ServiceControl.Operations.FailedMessageFactory.ParseFailureDetails(IReadOnlyDictionary`2 headers) in /_/src/ServiceControl/Operations/FailedMessageFactory.cs:line 41

   at ServiceControl.Operations.ErrorProcessor.ProcessMessage(MessageContext context, IIngestionUnitOfWork unitOfWork) in /_/src/ServiceControl/Operations/ErrorProcessor.cs:line 115

2026-04-11 08:58:18.9908|00:08:30.517|3|Warn|ServiceControl.Operations.ErrorIngestor|Processing of message '2b925812-3cba-4a95-b629-3688882ea9f3' failed|System.Exception: Missing 'NServiceBus.FailedQ' header.
```

andreasohlund:
@DavidBoike @danielmarbach @mattmercurio I've updated the issue description with more background details. Please do a final review

src/CommandLine/Queue.cs
```diff
@@ -2,23 +2,41 @@
...
+            if (forwardDeadLetteredMessagesTo.HasValue())
+            {
+                queueDescription.ForwardDeadLetteredMessagesTo = forwardDeadLetteredMessagesTo.Value().ToHierarchyNamespaceAwareDestination(hierarchyNamespace);
+                var deadLetterTargetQueueOptions = BuildDefaultCreateQueueOptions(queueDescription.ForwardDeadLetteredMessagesTo);
+
+                try
+                {
+                    await client.CreateQueueAsync(deadLetterTargetQueueOptions);
+                }
+                catch (ServiceBusException ex) when (ex.Reason == ServiceBusFailureReason.MessagingEntityAlreadyExists)
```
danielmarbach, line 24: I am not sure if it is a big deal and worth changing much. Just thinking out loud. It seems that with this change we now have the deal of creating one queue and the ramification that the existence of the queue in two places. Once here because we might need to create another queue and also one layer up because the line 30 queue creation assumes the ServiceBusException will be caught by the upper layer

src/CommandLineTests/CommandLineTests.cs
```diff
@@ -56,6 +57,34 @@ public async Task Create_hierarchy_namespace_endpoint()
...
+        public async Task Create_endpoint_with_dlq_forwarding()
+        {
+            await DeleteQueue(QueueName);
+            await DeleteQueue(ErrorQueueName);
+
+            var (output, error, exitCode) = await Execute($"endpoint create {EndpointName} --forward-dlq-to {ErrorQueueName}");
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(exitCode, Is.EqualTo(0));
```
danielmarbach: Isn't there an `Is.Zero`?

src/CommandLineTests/CommandLineTests.cs
```diff
@@ -56,6 +57,34 @@ public async Task Create_hierarchy_namespace_endpoint()
...
+                Assert.That(error, Is.EqualTo(string.Empty));
```
danielmarbach: Isn't there an `Is.Empty`?

danielmarbach: These comments apply to other such occurences

src/Transport/Receiving/DeadLetterRequest.cs
```diff
@@ -0,0 +1,39 @@
+namespace NServiceBus.Transport.AzureServiceBus;
+
+using System;
+using System.Collections.Generic;
+
+/// <summary>
+/// Requests the transport to dead-letter the message being processed when added to the transport transaction exposed to onError.
+/// </summary>
+public class DeadLetterRequest(string deadLetterReason, string deadLetterErrorDescription, Dictionary<string, object>? propertiesToModify = null)
```
danielmarbach: The `onError` callback receives a `TransportTransaction`, and in raw transport scenarios without the Core recoverability pipeline, that's the only viable entry point. Making `DeadLetterRequest` public enables those scenarios if I understand correctly.

I'd recommend documenting the two usage paths explicitly here:
1. With Core recoverability: Use `RecoverabilityAction.DeadLetter() `(the `DeadLetterRequest` is created and wired internally)
2. Raw transport: Set `DeadLetterRequest` directly in TransportTransaction via onError
This prevents users from accidentally combining both (e.g., returning `RecoverabilityAction.DeadLetter()` and setting a competing DeadLetterRequest in the transaction, where the second would silently overwrite the first). 

A brief `<remarks>` on here covering both paths and their precedence would close that gap.

src/Transport/Receiving/MessagePump.cs
```diff
@@ -283,6 +291,26 @@ await processMessageEventArgs.SafeCompleteMessage(message,
 
                     result = await onError!(errorContext, messageProcessingCancellationToken).ConfigureAwait(false);
 
+                    if (azureServiceBusTransaction.TransportTransaction.TryGet<DeadLetterRequest>(out var deadLetterRequest))
+                    {
+                        if (TransactionMode == TransportTransactionMode.None)
+                        {
+                            Logger.Error($"User requested message with id '{nativeMessageId}' to be moved to the dead-letter queue due to '{deadLetterRequest.DeadLetterReason}' but since transport transaction mode is 'None' dead lettering is not possible, message will be discarded");
```
danielmarbach: Is this an error given the user has chosen `None` which means `ReceiveAndDelete`?

src/Transport/Receiving/MessagePump.cs
```diff
@@ -138,7 +138,15 @@ async Task OnProcessMessage(ProcessMessageEventArgs arg)
...
         catch (Exception ex)
         {
-            await arg.SafeDeadLetterMessage(message, TransactionMode, ex, CancellationToken.None).ConfigureAwait(false);
+            if (TransactionMode == TransportTransactionMode.None)
+            {
+                Logger.Error($"Poison message detected.", ex);
+                return;
+            }
+
+            await arg.SafeDeadLetterMessage(message, nativeMessageId, new DeadLetterRequest(ex), CancellationToken.None).ConfigureAwait(false);
```
danielmarbach: I'm not sure what is better. Just thinking out loud. SafeXyz methods are highly pump-specific helpers that currently internally check some boundary conditions by checking the transport transaction mode and som other things. By introducing this change some utilities do require the transaction mode and some don't which creates some level of "mental" gap. Would it be better to move that check into the SafeDeadLetter message and only pass in the exception? Then below we could have a `TryDeadLetterIfNecessary` type of helper that checks for the dead letter request and the transaction mode and conditionally ignores or calls SafeDeadLetter

src/Transport/Receiving/FaultsDeadLetteringFeature.cs
```diff
@@ -0,0 +1,34 @@
...
+using ConsistencyGuarantees;
+using Features;
+using Pipeline;
+using Transport.AzureServiceBus;
+
+class FaultsDeadLetteringFeature : Feature
+{
+    protected override void Setup(FeatureConfigurationContext context)
+    {
+        if (context.Settings.GetRequiredTransactionModeForReceives() == TransportTransactionMode.None)
```
danielmarbach, line 14: The pump code guards against TransportTransactionMode.None and then logs but this code makes sure we cannot run in a mode where we would ever try to deadletter with ReceiveAndDelete. I think throwing here is fine but if we throw here it should not be an argument exception because there is no argument the user provided here. Should it be a general exception or an InvalidOperationException?

Now seeing this it probably means this is the justification to "log.Error" in the other location I commented on but I still think if that really makes sense given with ReceiveAndDelete it is impossible for us to deadletter because the user has chosen an ACK on receive mode.

src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -35,6 +36,21 @@ static class MessageExtensions
             headers[Headers.ContentType] = message.ContentType;
         }
 
+        if (!string.IsNullOrWhiteSpace(message.DeadLetterSource) && !headers.ContainsKey(FaultsHeaderKeys.FailedQ))
```
danielmarbach: I don't know if it matters, but intuitively I would have flipped it around to first check the absence of the FailedQ header because we want to express that we are never overriding those headers when they are present.

The same applies for other conditions

src/Transport/Receiving/FaultsDeadLetteringFeature.cs
danielmarbach: Another option is that we could validate incorrect configuration directly in the transport initialize or in the property setters of the transport instead of in the feature because that would mean the validation would always apply even in the raw mode.

src/Transport/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -73,6 +73,7 @@ void WriteStartupDiagnostics(StartupDiagnosticEntries startupDiagnostic) =>
             CustomTokenProvider = transportSettings.TokenCredential?.ToString() ?? "default",
             CustomRetryPolicy = transportSettings.RetryPolicyOptions?.ToString() ?? "default",
             AutoDeleteOnIdle = transportSettings.AutoDeleteOnIdle?.ToString() ?? "default",
+            AutoForwardDeadLetteredMessagesToErrorQueue = transportSettings.AutoForwardDeadLetteredMessagesToErrorQueue.ToString(),
```
danielmarbach, line 76: Can we ever be in a situation where we have this set to true but no error queue? I think not because we always have a default error queue but I wanted to double check

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -184,6 +171,52 @@ void ApplyWebProxyIfNeeded(ServiceBusClientOptions options)
...
+            {
+                receiveQueue.ForwardDeadLetteredMessagesTo = errorQueueName;
+            }
+
+            queues[receiveQueueName] = receiveQueue;
+
+            queues[errorQueueName] = BuildDefaultCreateQueueOptions(errorQueueName);
+        }
+
+        // make sure that queues without forwarding are created first.
```
danielmarbach: Should we add a comment why? Is my understanding correct?

This ensures non-forwarding queues are created first. ASB requires the target of a forward to exist before the source. But the error queue is included in sendingAddresses (added first) AND also re-added in the receiver loop. Since queues is a dictionary with case-insensitive comparer, the error queue from sendingAddresses gets overwritten by the one from the receiver loop (which also uses `BuildDefaultCreateQueueOptions`). This is fine since both produce identical options, but it means the error queue will never have `ForwardDeadLetteredMessagesTo` set on it (it's built with `BuildDefaultCreateQueueOptions` only). This is by design to prevent self-forwarding loops

src/Transport/Receiving/MessageExtensions.cs
danielmarbach: Could we ever corner ourselves into a situation where someone uses the `RecoverabilityAction.DeadLetter(...)` overload because of the presence of those headers that override doesn't take precedence?

src/Transport/Receiving/RecoverabilityActionExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+/// Provides Azure Service Bus-specific recoverability actions.
+/// </summary>
+public static class RecoverabilityActionExtensions
+{
+    extension(RecoverabilityAction _)
+    {
+        /// <summary>
+        /// Creates a recoverability action that moves the message to the dead-letter queue with the specified details.
+        /// </summary>
+        public static DeadLetterMessage DeadLetter(string deadLetterReason, string deadLetterErrorDescription, Dictionary<string, object>? propertiesToModify = null)
```
danielmarbach: Should this be `IReadOnlyDictionary<string, object>?` similar to what the SDK accepts? This doesn't seem to be super perf sensitive and therefore having the interface might make sense

src/Transport/Receiving/RecoverabilitySettingsExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+/// Extension to the core RecoverabilitySettings.
+/// </summary>
+public static class RecoverabilitySettingsExtensions
+{
+    extension(RecoverabilitySettings settings)
+    {
+        /// <summary>
+        /// Uses the endpoints native service bus dead letter queue for message failures.
+        /// </summary>
+        public void MoveErrorsToAzureServiceBusDeadLetterQueue() => settings.GetSettings().EnableFeature<FaultsDeadLetteringFeature>();
```
danielmarbach, line 16: Isn't this wrong because it effectively means you can never use a custom recoverability policy that returns the dead letter message action without calling `MoveErrorsToAzureServiceBusDeadLetterQueue`. I think the transport has to always enable the feature but the feature can disable itself if the transport definition has the flag not set.

src/Transport/Receiving/DeadLetterRequest.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// Dead letters the message being processed due to the provided exception.
+    /// </summary>
+    public DeadLetterRequest(Exception exception, Dictionary<string, object>? propertiesToModify = null) : this(
+        $"{exception.GetType().FullName}{Separator}{exception.Message}",
+        exception.StackTrace ?? "No stack trace available",
+        propertiesToModify)
+    {
+    }
+
+    static string Truncate(string value, int maxLength) => value.Length <= maxLength ? value : value[..maxLength];
```
danielmarbach: This silently truncates without any indication (no ellipsis, no log warning). Users may not realize their reason/description was cut. Is that OK?

src/Transport/Receiving/RecoverabilitySettingsExtensions.cs
andreasohlund: Discussed and since this only flips the default means that .DeadLetter() can still be used so keeping this as is

src/Transport/AzureServiceBusTransportInfrastructure.cs
andreasohlund: Correct, core will always default to "error"

src/Transport/AzureServiceBusTransport.cs
andreasohlund: @danielmarbach pushed https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1373/changes/7d056e813f95927876c8817391c2a26d7fa86284 to add more comments. This also makes the code more strict with Add and TryAdd, not sure if that is a good thing, eg if core now would add duplicate sending addresses we would blow up?

src/Transport/Receiving/RecoverabilityActionExtensions.cs
andreasohlund: DeadLetterMessageAsync uses IDictionary so went with that

src/Transport/Receiving/MessageExtensions.cs
andreasohlund: Talked to Daniel and we agreed that the order should be changed, I will also make sure that we have test coverage for this

src/CommandLine/Queue.cs
andreasohlund: I tried to find a clean way to do this wasn't able so I propose to leave as is

danielmarbach (approved):

PR #1396, opened 2026-04-20
Throwing exception on publish to a missing topic

### LINKED ISSUE #1255: Add option ThrowOnMissingTopicWhenPublishing to fail message publishing when configured topic is missing in the topology
#### Is your feature related to a problem?

When using Azure Service Bus and the new topology, we do not get an error when the topic for an event does not exist. In fact the exception is caught and swallowed in [MessageDispatcher](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/Sending/MessageDispatcher.cs).

While we understand that [this is by design](https://discuss.particular.net/t/publish-event-no-exception-when-azure-service-bus-topic-does-not-exist/4441), this has some downside for us:

* if we want to deploy topics and subscriptions separately, e.g. via Terraform, we get only limited feedback about mistakes, for instance when we forgot to add a topic
* if a topic disappears for any other reason (which may include accidential manual modification), we might not get instant feedback
* attempting to publish to an ASB topic that does not exist is pretty expensive/slow and includes retries, which affects other topics and queues too under higher load, see #00100320 in your internal support ticketing system

#### Describe the requested feature

We'd like a way to get instant feedback at runtime when we try to publish to a topic that does not exists in Azure Service Bus (whether or not anyone subscribes to it), either by:
* bubbling up the Exception (or throwing a new one)
* any interception mechanism so that we can throw an Exception ourselves or write error logs
        
#### Describe alternatives you've considered

* our Terraform plans check (obviously) that topics exist when we subscribe to them
* this does not solve the issue about events that are published but not subscribed (yet)
* we thought about a "never publish an event no one subscribes to", but we need this at least temporarily, so we can introduce an event and then let another team build their handlers

### Additional Context

- https://discuss.particular.net/t/publish-event-no-exception-when-azure-service-bus-topic-does-not-exist

afprtclr:
Fix https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/1255

This PR adds a new `ThrowOnMissingTopicWhenPublishing` transport configuration property, set to `false` by default for backward compatibility, that, when set to `true`, will cause the transport to throw an `InvalidOperationException` when publishing to a non-existent topic.

Alongside that, the message dispatcher now always logs a warning when publishing to a non-existent topic.

### DIFF
--- a/src/Tests/Sending/MessageDispatcherTests.cs
+++ b/src/Tests/Sending/MessageDispatcherTests.cs
@@ -148,6 +148,43 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Sending
             Assert.That(async () => await dispatcher.Dispatch(new TransportOperations(operation1, operation2), new TransportTransaction()), Throws.Nothing);
         }
 
+        // With pub sub the cases of the topic not being available are similar to having a topic without any subscribers
+        [Test]
+        public void Should_throw_when_multicast_dispatch_destination_not_available_and_throw_on_missing_topic_enabled()
+        {
+            var client = new FakeServiceBusClient();
+
+            var dispatcher = new MessageDispatcher(client, new MessageSenderRegistry(), TopicTopology.FromOptions(new TopologyOptions()
+            {
+                PublishedEventToTopicsMap = { { typeof(SomeEvent).FullName, "sometopic" } }
+            }), throwOnMissingTopic: true);
+
+            var sender = new FakeSender
+            {
+                SendMessageAction = _ => throw new ServiceBusException("Some exception", ServiceBusFailureReason.MessagingEntityNotFound),
+                SendMessageBatchAction = _ => throw new ServiceBusException("Some exception", ServiceBusFailureReason.MessagingEntityNotFound)
+            };
+            client.Senders["sometopic"] = sender;
+
+            var operation1 =
+                new TransportOperation(new OutgoingMessage("SomeId",
+                        [],
+                        ReadOnlyMemory<byte>.Empty),
+                    new MulticastAddressTag(typeof(SomeEvent)),
+                    [],
+                    DispatchConsistency.Isolated);
+
+            var operation2 =
+                new TransportOperation(new OutgoingMessage("SomeId",
+                        [],
+                        ReadOnlyMemory<byte>.Empty),
+                    new MulticastAddressTag(typeof(SomeEvent)),
+                    [],
+                    DispatchConsistency.Default);
+
+            Assert.That(async () => await dispatcher.Dispatch(new TransportOperations(operation1, operation2), new TransportTransaction()), Throws.InvalidOperationException);
+        }
+
         [Test]
         public async Task Should_dispatch_unicast_isolated_dispatches_individually_per_destination()
         {
@@ -744,7 +781,7 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Sending
             var client = new FakeServiceBusClient();
             var destinationManager = GetDestinationManager("SomeHierarchyNamespace");
 
-            var dispatcher = new MessageDispatcher(client, new MessageSenderRegistry(), TopicTopology.FromOptions(new TopologyOptions()), destinationManager);
+            var dispatcher = new MessageDispatcher(client, new MessageSenderRegistry(), TopicTopology.FromOptions(new TopologyOptions()), destinationManager, throwOnMissingTopic: false);
 
             var operation1 =
                 new TransportOperation(new OutgoingMessage("SomeId",
@@ -802,7 +839,7 @@ namespace NServiceBus.Transport.AzureServiceBus.Tests.Sending
                 {
                     PublishedEventToTopicsMap = { { typeof(SomeEvent).FullName, "sometopic" } }
                 }),
-                destinationManager);
+                destinationManager, throwOnMissingTopic: false);
 
             var operation1 =
                 new TransportOperation(new OutgoingMessage("SomeId",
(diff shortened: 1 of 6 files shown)


src/Transport/Sending/MessageDispatcher.cs
```diff
@@ -230,9 +232,14 @@ async Task DispatchBatchOrFallbackToIndividualSendsForDestination(string destina
...
-                if (Log.IsDebugEnabled)
+                if (Log.IsWarnEnabled)
+                {
+                    Log.Warn($"Skipping sending messages to topic {destination} because the destination does not exist.");
+                }
+
+                if (throwOnMissingTopic)
                 {
-                    Log.Debug($"Skipping sending messages to topic {destination} because the destination does not exist.");
+                    throw new InvalidOperationException($"Sending messages to topic {destination} failed because the destination does not exist.", e);
```
mauroservienti: Should we be more specific? (in the log warn too?)

```suggestion
                    throw new InvalidOperationException($"Publishing messages to topic {destination} failed because the destination does not exist.", e);
```

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -450,6 +450,11 @@ public IWebProxy? WebProxy
     /// </remarks>
     public OutgoingNativeMessageCustomizationAction? OutgoingNativeMessageCustomization { get; set; }
 
+    /// <summary>
+    /// Specifies whether to throw an exception when publishing to a non-existent topic
+    /// </summary>
+    public bool ThrowOnMissingTopicWhenPublishing { get; init; }
```
mauroservienti: Considering other properties are get/set, what are the implications of `init` here?

afprtclr: `init` is like `set` that one can use only during object initialization.

Since this is propagated down the line and things are configured on initialization, this is effectively read once.

src/Transport/Sending/MessageDispatcher.cs
afprtclr: Fixed.

danielmarbach:
FYI we are planning to release 6.3.0 in the not too-distant future and we can bundle this if you want

src/Transport/AzureServiceBusTransport.cs
mauroservienti: Does it mean that users must use object initializers? and the following won't work:

```csharp
var transport = new AzureServiceBusTransport();
transport.ThrowOnMissingTopicWhenPublishing = true; // this won't compile
```

and users are forced to:

```csharp
var transport = new AzureServiceBusTransport()
{
  transport.ThrowOnMissingTopicWhenPublishing = true
}
```

danielmarbach: FYI Transport Extensions also need to be changed which most of the time means the property cannot be init only :(

We are adding a convention test now to make sure those things are aligned because this "guideline in platform" wasn't followed https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1397

afprtclr:
> FYI we are planning to release 6.3.0 in the not too-distant future and we can bundle this if you want

Love it! This is the only work item we plan in ASB so far, so we'll appreciate if that can go in one batch. We'll have one more item in ServiceControl + another item in Azure Functions, so we can coordinate these releases as well if you plan on doing them.

src/Transport/AzureServiceBusTransport.cs
afprtclr: > Does it mean that users must use object initializers? and the following won't work:

Yes, that is correct.

> FYI Transport Extensions also need to be changed which most of the time means the property cannot be init only :(

Right, I can see that now. I'll change the property to `set` and add an extension method to enable/disable throwing this exception. Thanks for catching that.

mauroservienti: @afprtclr, I pushed  https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1396/commits/b61f5725ea3945ff9cd4691c7a17104c02d81291 to address it

mauroservienti (approved):
This is good to go, IMO. We also need to document the new setting.

src/Tests/Sending/MessageDispatcherTests.cs
```diff
@@ -148,6 +148,43 @@ public void Should_swallow_when_multicast_dispatch_destination_not_available()
             Assert.That(async () => await dispatcher.Dispatch(new TransportOperations(operation1, operation2), new TransportTransaction()), Throws.Nothing);
         }
 
+        // With pub sub the cases of the topic not being available are similar to having a topic without any subscribers
+        [Test]
+        public void Should_throw_when_multicast_dispatch_destination_not_available_and_throw_on_missing_topic_enabled()
```
cquirosj, line 153: We can polish this test so we take out the multicast aspect.

src/Tests/Sending/MessageDispatcherTests.cs
```diff
@@ -660,6 +697,78 @@ await dispatcher.Dispatch(new TransportOperations([.. operations]),
...
+            var azureServiceBusTransaction = new AzureServiceBusTransportTransaction();
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(async () => await dispatcher.Dispatch(new TransportOperations([.. operations]), azureServiceBusTransaction.TransportTransaction), Throws.InvalidOperationException);
+                Assert.That(defaultSender.BatchSentMessages, Has.Count.Zero);
+                Assert.That(defaultSender.IndividuallySentMessages, Has.Count.EqualTo(0));
+            });
+        }
+
```
cquirosj, line 771: Same as my other comment. We explore making these tests easier to read by removing the multicast part.

src/Tests/Sending/MessageDispatcherTests.cs
afprtclr: Not sure what you mean by this exactly. The thing we're fixing is only for multicast. Specifically, we shouldn't throw on unicast (or at least haven't decided to do so).

mauroservienti: The transport already blows up on unicast if the destination queue doesn't exist. This check

https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/7d824335c94818eddc82cee3d4621d68162385a1/src/Transport/Sending/MessageDispatcher.cs#L231-L239

only "hides" multicast failures, and that's what we are changing

cquirosj (approved):

PR #1397, opened 2026-04-21
Align transport extension API with newly added transport properties

danielmarbach:
This PR catches up with a convention enforcement test to make sure this guideline is not missed.

> However, since the API is not yet marked as obsolete, and hence users are not seeing warnings about using this API, any new transport settings should always be added to both.

Ideally this convention would exist in Core, but that is something we can do as a next step because I have a hard time also envisioning what such a generic test would look like given the `PreObsolete` information is erased from the API surface.


danielmarbach:
FYI @jpalac @warwickschroeder @PhilBastian @abparticular @ngallegos

src/Tests/Configuration/TransportSettingsConventionTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        var transportType = typeof(AzureServiceBusTransport);
+
+        var transportProperties = transportType
+            .GetProperties(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly)
+            .Where(p => !IsObsolete(p))
+            .Where(p => p.SetMethod is not null && p.SetMethod.IsPublic)
+            .ToList();
+
+        var extensionSource = File.ReadAllText(ResolveTransportExtensionSourcePath());
```
danielmarbach, line 28:
```suggestion
        // We are reading the source file instead of using reflection because PreObsolete metadata is conditionally compiled away
        var extensionSource = File.ReadAllText(ResolveTransportExtensionSourcePath());
```

src/Tests/ApprovalFiles/TransportSettingsConventionTests.Transport_properties_have_PreObsolete_extension_method_mappings.approved.txt
```diff
@@ -0,0 +1,16 @@
...
+  - MaxAutoLockRenewalDuration
+  - PrefetchCount
+  - PrefetchMultiplier
+  - RetryPolicyOptions
+  - TimeToWaitBeforeTriggeringCircuitBreaker
+  - UseWebSockets
+
+Unmapped properties:
+  - OutgoingNativeMessageCustomization
+  - WebProxy
```
danielmarbach, line 16: WebProxy is an optional property on the `UseWebSocket` extension method and therefore technically is mapped, but I didn't want to further complicate the test infrastructure.

src/Tests/ApprovalFiles/TransportSettingsConventionTests.Transport_properties_have_PreObsolete_extension_method_mappings.approved.txt
```diff
@@ -0,0 +1,16 @@
...
+  - HierarchyNamespaceOptions
+  - MaxAutoLockRenewalDuration
+  - PrefetchCount
+  - PrefetchMultiplier
+  - RetryPolicyOptions
+  - TimeToWaitBeforeTriggeringCircuitBreaker
+  - UseWebSockets
+
+Unmapped properties:
+  - OutgoingNativeMessageCustomization
```
danielmarbach: My understanding is this one is technically an ultra low level seam introduced for MT but given this is now also used and documented in the advanced multiplexing scenarios, we probably should bite the bullet and add a corresponding extension method too?

andreasohlund: Agreed

src/Transport/AzureServiceBusTransportSettingsExtensions.cs
```diff
@@ -52,6 +53,25 @@ public static TransportExtensions<AzureServiceBusTransport> UseTransport<TTransp
...
+    /// </summary>
+    /// <param name="transportExtensions"></param>
+    /// <param name="autoDeleteOnIdle">The idle timeout after which unused entities are automatically deleted. The minimum allowed value is 5 minutes.</param>
+    /// <remarks>
+    /// This setting only applies to queues, not to topics or subscriptions. Topics and subscriptions are considered
+    /// shared infrastructure and are not affected by this property. Only instance-specific input queues (such when using 'MakeInstanceUniquelyAddressable') 
+    /// will have AutoDeleteOnIdle applied, while shared queues (such as error and audit queues) remain unaffected to prevent unintended deletion of critical infrastructure.
+    /// </remarks>
+    [PreObsolete("https://github.com/Particular/NServiceBus/issues/6811",
+        Note = "Should not be converted to an ObsoleteEx until API mismatch described in issue is resolved.",
```
andreasohlund:
```suggestion
        Note = "Should not be converted to an ObsoleteEx until API mismatch described in the issue is resolved.",
```
Nitpick...

Should be applied to the other notes as well if pulled in

andreasohlund (approved):

DavidBoike (approved):

Issue #1337, opened 2026-03-02
Azure Service Bus Transport: Auto-provision correlation filters for both publish and subscribe when events are multiplexed to a shared topic


danielmarbach:
When using `topology.PublishTo<TEvent>("SharedTopic")` to multiplex multiple concrete event types through a single topic, the transport currently has no awareness of this multiplexing at install time. Users must manually provision subscription rules correlation filters) via IaC or the `ServiceBusAdministrationClient` for both the publishing side (stamping the concrete type) and the subscribing side (filtering to only the desired type). 

It would be a significant improvement if the transport could detect this multiplexing pattern and automatically provision sensible correlation filter rules on both sides when endpoint installers are enabled — but this should be an explicit opt-in, not a default behaviour.

## Motivation

A common use case for topic multiplexing is polymorphic event hierarchies where:

- A base event type (e.g. `ProductOrdered`) is a stable contract in its own assembly
- Concrete derived types (e.g. `CarInsuranceProductOrdered`) are published to the  shared base topic via `PublishTo<TConcrete>("ProductOrdered")`
- Two kinds of subscribers exist:
  - Endpoints handling only a specific concrete type — these need a correlation 
    filter to avoid receiving all events on that topic
  - Endpoints handling the base type — these want everything on the topic and need 
    a catch-all (no filter) subscription

Without automatic provisioning, users must step outside the NServiceBus installer pipeline to configure these rules, splitting infrastructure concerns across multiple tools and creating a sharp edge when onboarding new event types.

## Context

Topic multiplexing and polymorphic routing are advanced, deliberate topology choices. The recommended default for most systems remains topic-per-event, which avoids filtering complexity entirely and maps cleanly to NServiceBus's event model. 

Any installer support for correlation filters should therefore be:

- **Explicitly opted into** via topology configuration, not inferred automatically   from the presence of `PublishTo` overrides
- **Clearly documented as an advanced scenario** with guidance on when it is and isn't appropriate
- **Not a replacement** for the auto-forwarding pattern, which remains the better fit when subscribers cannot reference concrete types at all

The goal is to reduce friction for teams who have consciously chosen multiplexing, not to make multiplexing feel like a first-class default path.

### Publish side

[...]

PR #1399, opened 2026-04-22
Multiplexed shared-topic filter support for the default topology

danielmarbach:
Addresses:

- #1337 

This PR introduces **multiplexed shared-topic filter support** for Azure Service Bus while preserving the transport's recommended default model of **topic-per-event**.

The feature is intentionally designed for advanced scenarios where multiple event types need to share a topic, without forcing that complexity onto the default path. It adds first-class support for:

- Explicit shared-topic routing
- CorrelationFilter-based multiplexing
- SqlFilter-based multiplexing
- Fallback-topic based polymorphic subscriptions for unmapped event types

The result is a simpler and more extensible routing model that supports both deliberate topic sharing and pragmatic subscription ergonomics.

## Design goals

The implementation keeps the default guidance unchanged:

1. Prefer **topic-per-event**
2. If multiplexing is required, prefer **CorrelationFilter**
3. Use **SqlFilter** only when correlation filtering is insufficient

This ensures the common case remains simple, while advanced topologies become easier to model.

### Unified routing model

Instead of separate publisher/subscriber mode systems, the PR consolidates behavior into a single routing concept:

`TopicRoutingMode`

Supported modes:

- `NotMultiplexed`
- `CorrelationFilter`
- `SqlFilter`

This reduces API surface area and removes the need for publisher/subscriber compatibility matrices.

### Explicit opt-in shared-topic routing

Multiplexing is never accidental.

Users must explicitly choose shared-topic behavior through routing configuration. If they do nothing, existing topic-per-event behavior remains intact.

That keeps operational intent visible and avoids hidden coupling.

### Fallback topic support for unmapped events

A new `FallbackTopic` concept allows **unmapped** event types to route through a shared topic.

Events explicitly mapped to `FallbackTopic.TopicName` without a per-event routing mode inherit the fallback mode as well, keeping publisher stamping and subscriber rules symmetric across both regular and hierarchy-prefixed setups.

This is particularly valuable for subscribers using:

- base contracts
- interfaces
- polymorphic event hierarchies

Without requiring explicit mappings for every concrete event type.

That means teams can keep topic sharing deliberate while avoiding brittle per-event subscription maintenance.

### CorrelationFilter mode

Publisher dispatch stamps normalized enclosed message types as boolean application properties.

Subscribers provision Azure Service Bus correlation rules that match exact type identities.

Benefits:

- precise matching
- fast native filtering
- avoids string scanning
- supports significantly higher rule counts than SQL filters
- scales better for large multiplexed topologies

This makes correlation filters the preferred mode whenever possible.

### SqlFilter mode

Uses SQL rules over `NServiceBus.EnclosedMessageTypes`, preserving historical behavior and compatibility.

Benefits:

- leverages existing metadata
- useful when correlation filters are insufficient
- supports scenarios requiring expression-based matching

Tradeoffs:

- lower rule-count limits than correlation filters
- less precise than correlation filters
- substring matching can overmatch similarly named types

## Validation changes

The implementation simplifies validation rules.

`FallbackTopic.TopicName` and `FallbackTopic.Mode` are validated declaratively, so misconfigurations surface at startup rather than at first publish. A hierarchy-aware post-prefix length check catches names that exceed topic-name limits only after the prefix is applied.

`NotMultiplexed` and `CorrelationFilter` can coexist on the same topic/subscription because Azure Service Bus rules are OR-based. A catch-all rule already receives everything, making additional filtered rules redundant but harmless.

`SqlFilter` remains distinct because it represents a different filtering strategy.

### Why no custom filter override API?

This PR intentionally does **not** introduce a callback-based "bring your own filter" model.

That path creates a synchronization problem:

- publishers must stamp exactly what subscribers expect
- subscribers must define filters matching publisher conventions

This reintroduces complexity the new unified design removes.

For advanced one-off scenarios, users can already use existing native message customization hooks.

### Why CorrelationFilter over SqlFilter?

Correlation filters are preferred because they are:

- explicit
- deterministic
- contract-oriented
- operationally efficient
- able to scale to larger numbers of rules

SQL filters remain available for compatibility and scenarios correlation filters cannot express.

## Future extensibility

This design deliberately leaves room for future enhancements without reshaping the public model.

### SqlFilterPrecise

A stricter SQL matching mode that respects token boundaries instead of substring matching, avoiding accidental matches like:

- `OrderAccepted`
- `OrderAcceptedV2`

### TopicRoutingMode.Custom

A future advanced mode could allow custom publisher stamping and custom rule creation for the small percentage of highly specialized systems.

Because the current architecture centralizes dispatch stamping and subscription rule creation, such additions are feasible without redesigning the model.

## Why this matters

This PR gives users the best of both worlds:

- Keep topic-per-event as the recommended default
- Support shared-topic routing when intentionally needed
- Prefer the filter strategy that scales best operationally
- Enable polymorphic subscription scenarios with less manual mapping
- Keep the public API compact and understandable
- Preserve room for future routing/filter innovations

In short, it adds advanced topology flexibility **without making the common case more complex**.


src/AcceptanceTests/Subscribing/When_using_topic_per_event_topology_with_selective_correlation_filter_multiplexing.cs
```diff
@@ -0,0 +1,150 @@
...
+            Assert.That(context.Subscriber1GotMyEvent2, Is.False);
+            Assert.That(context.Subscriber2GotMyEvent1, Is.False);
+            Assert.That(context.Subscriber2GotMyEvent2, Is.True);
+        }
+    }
+
+    public class Context : ScenarioContext
+    {
+        public bool Subscriber1GotMyEvent1 { get; set; }
+        public bool Subscriber1GotMyEvent2 { get; set; }
```
mauroservienti: This is never set, and the assert above will alwyas be true

src/AcceptanceTests/Subscribing/When_using_topic_per_event_topology_with_selective_correlation_filter_multiplexing.cs
```diff
@@ -0,0 +1,150 @@
...
+        public bool Subscriber2GotMyEvent1 { get; set; }
```
mauroservienti: This is never set, and the assert above will alwyas be true

src/AcceptanceTests/Subscribing/When_using_topic_per_event_topology_with_selective_mapped_and_fallback_correlation_filter_on_shared_topic.cs
```diff
@@ -0,0 +1,166 @@
...
+            Assert.That(context.Subscriber1GotMyEvent2, Is.False);
+            Assert.That(context.Subscriber2GotMyEvent1, Is.False);
+            Assert.That(context.Subscriber2GotMyEvent2, Is.True);
+        }
+    }
+
+    public class Context : ScenarioContext
+    {
+        public bool Subscriber1GotMyEvent1 { get; set; }
+        public bool Subscriber1GotMyEvent2 { get; set; }
```
mauroservienti: This is never set, and the assert above will alwyas be true

src/AcceptanceTests/Subscribing/When_using_topic_per_event_topology_with_selective_mapped_and_fallback_correlation_filter_on_shared_topic.cs
```diff
@@ -0,0 +1,166 @@
...
+        public bool Subscriber2GotMyEvent1 { get; set; }
```
mauroservienti: This is never set, and the assert above will alwyas be true

src/AcceptanceTests/Subscribing/When_using_topic_per_event_topology_with_selective_mapped_and_fallback_correlation_filter_on_shared_topic.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                var endpointName = Conventions.EndpointNamingConvention(typeof(SubscriberForEvent1));
+                var topology = (TopicPerEventTopology)TopicTopology.FromOptions(new TopologyOptions
+                {
+                    FallbackTopic = new FallbackTopicOptions
+                    {
+                        TopicName = SharedTopicName,
+                        Mode = TopicRoutingMode.CorrelationFilter
+                    }
+                });
```
mauroservienti: Is my understanding correct that for this test case the fallback configuration for SubscriberForEvent1 is not needed? Publisher publishes MyEvent1 to the shared topic and this subscriber is explicitely subscribed to that type.

src/Transport/EventRouting/TopicRoutingMode.cs
```diff
@@ -0,0 +1,50 @@
+namespace NServiceBus.Transport.AzureServiceBus;
+
+/// <summary>
+/// Controls how a shared topic is used for publishing and subscribing.
```
SzymonPobiega:
```suggestion
/// Controls how the topic is used for publishing and subscribing.
```

src/Transport/EventRouting/TopicRoutingMode.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// On unsubscribe, deletes the matching subscription rule.
+    /// </summary>
+    SqlFilter,
+
+    /// <summary>
+    /// Provisions a plain topic subscription with no filter rule so all messages on the topic are forwarded.
+    /// This mode does not add any extra publish-side message data beyond the normal outgoing message headers.
+    /// On unsubscribe, deletes that subscription.
+    /// </summary>
+    CatchAll
```
SzymonPobiega: What is the purpose of that? The description sounds like `NotMultiplexed` unless I am missing something. I also looked for references and the only one I found was in `ValidateFallbackTopic`.

src/Transport/EventRouting/TopicRoutingMode.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <summary>
+    /// Defers to the effective topology resolution rules.
+    /// For publishing, this resolves to per-event <see cref="RoutingOptions.Mode"/> when set,
+    /// otherwise to <see cref="FallbackTopicOptions.Mode"/> for unmapped events when a fallback topic is configured,
+    /// otherwise to <see cref="NotMultiplexed"/>.
+    /// For subscribing, this resolves to the explicitly configured <see cref="SubscriptionEntry.RoutingMode"/> when set,
+    /// otherwise to <see cref="FallbackTopicOptions.Mode"/> for unmapped subscriptions when a fallback topic is configured,
+    /// otherwise to <see cref="CatchAll"/>.
+    /// </summary>
+    Default,
```
SzymonPobiega: When is it legal to specify `Default` as the routing mode? It seems to be illegal in the fallback topic specification. It also seems to be not OK when overriding the routing mode for specific `PublishTo` / `Subscribe` instruction. 

Are we conflating two concepts here:
- as far as user-facing API is concerned, there are only 3 valid topic routing modes: `NotMultiplexed`, `CorrelationFilter` and `SqlFilter`.
- in the internal representation, we add `Default` as a null-like value that means "the user did not specify".

src/Transport/EventRouting/TopicRoutingMode.cs
danielmarbach: They are functionally identical in their subscription-side behavior, both create a plain subscription with no filter rules where NotMultiplexed and CatchAll fall through to the same CreateCatchAllSubscription call). 

The difference is semantic intent and resolution defaults:

1. NotMultiplexed means "I'm deliberately publishing to a topic without multiplexing" it's the resolved default on the publish side (unmapped publish events resolve to NotMultiplexed).
2. CatchAll means "I want to receive everything on this shared topic" it's the resolved default on the subscribe side (unmapped subscriptions resolve to CatchAll).
3. Validation compatibility: NotMultiplexed, CatchAll, and CorrelationFilter are all treated as compatible on the same topic/subscription because ASB OR-semantics makes a catch-all + correlation filter combination harmless.
4. FallbackTopic validation: CatchAll and NotMultiplexed are rejected as fallback modes  only CorrelationFilter or SqlFilter are valid for the fallback.

So CatchAll exists primarily to give subscribers an explicit way to say "subscribe to everything on this topic" that's semantically distinct from NotMultiplexed's "not using multiplexing at all." The infrastructure behavior is the same, but the enum separation preserves the publish vs. subscribe intent distinction and allows the resolution defaults to diverge (publish defaults to NotMultiplexed, subscribe defaults to CatchAll).

Do you think that is overkill?

src/Transport/EventRouting/TopicRoutingMode.cs
danielmarbach: Yeah that is confusing. The user shouldn't pick that as an option. I'll remove it from the public API surface.

danielmarbach: Done

src/Transport/EventRouting/TopicRoutingMode.cs
danielmarbach: Thinking about this further, I think it is overkill. With the last change, null already resolves on Publish to NotMultiplexes and subscribe resolves to CatchAll. Users who explicitly set NotMultiplexed on the subscription entry get the same behavior as CatchAll. The distinction doesn't matter at runtime. The normalization logic already collapses those two paths, and having CatchAll on the public API means we have to explain and document it. 

The only slight awkwardness is 

NotMultiplexed is a confusing name for a subscribe-side mode "I want all messages" doesn't naturally map to "not multiplexed." CatchAll reads better in subscription config. But that's a naming problem, not a behavioral one, and it could be solved with XML doc clarification rather than a redundant enum value.

danielmarbach:
@SzymonPobiega I think this is ready for a final review except for the missing rename of the enum that I forgot to push 🤦 Will push this tomorrow

mauroservienti (approved):

danielmarbach:
@mauroservienti I think I have addressed the comments. It is not entirely perfect, but I turned these tests into tests that do not automatically subscribe but have the second handler with MarkAsFailed. In case the filter population would be broken, those should fail. The issue is that the order is not guaranteed, so we might run into some flakiness, but I guess there is no better way at this stage.

src/AcceptanceTests/Subscribing/When_using_topic_per_event_topology_with_selective_mapped_and_fallback_correlation_filter_on_shared_topic.cs
danielmarbach: Correct

SzymonPobiega (approved):

PR #1400, opened 2026-04-22
Allow configuring max delivery count

danielmarbach:
This change exposes the `MaxDeliveryCount` setting directly on `AzureServiceBusTransport`, giving users explicit control over the maximum delivery count applied to queues created during endpoint infrastructure setup.

## Rationale

Previously, the transport used `int.MaxValue` as the hard-coded default for `MaxDeliveryCount`, which effectively disabled Azure Service Bus's built-in delivery count limits and deferred all retry decisions to NServiceBus recoverability. While this maintains backward compatibility, it prevents users from surfacing Azure Service Bus's native retry behavior and from tuning delivery count independently of recoverability policy.

By exposing `MaxDeliveryCount`, we:
- Make the setting explicit and discoverable in code-based configuration.
- Provide a path to change the default to a saner value in the next major version. The current default remains `int.MaxValue` to avoid breaking existing deployments.
- Allow the Azure Service Bus Emulator (which requires a value of `10`) to receive a sensible default automatically, while still permitting users to override it.

## Usage

```csharp
var transport = new AzureServiceBusTransport(connectionString, TopicTopology.Default)
{
    MaxDeliveryCount = 15
};
```

## Recommendations

When choosing a value, consider the interaction with NServiceBus recoverability:
- The value should be **high enough** to allow the configured recoverability policy to eventually move the message to the error queue.
- It should **not be so high** that it effectively creates infinite retries.
- If you prefer Azure Service Bus's native dead-letter queue over NServiceBus's error queue, enable `AutoForwardDeadLetteredMessagesToErrorQueue`.

### DIFF
--- a/src/Transport/AzureServiceBusTransport.cs
+++ b/src/Transport/AzureServiceBusTransport.cs
@@ -107,7 +107,7 @@ public partial class AzureServiceBusTransport : TransportDefinition
             ? new ServiceBusClient(FullyQualifiedNamespace, TokenCredential, defaultClientOptions)
             : new ServiceBusClient(ConnectionString, defaultClientOptions);
 
-        var administrationConnectionString = IsUsingDevelopmentEmulator
+        var administrationConnectionString = IsUsingDevelopmentEmulator(ConnectionString)
             ? InjectEmulatorAdminPort(ConnectionString!)
             : ConnectionString!;
         var administrationClient = TokenCredential != null
@@ -187,7 +187,7 @@ public partial class AzureServiceBusTransport : TransportDefinition
             // and core might allow different error queues in the future so we will not make any assumptions here
             var errorQueueName = DestinationManager.GetDestination(receiver.ErrorQueue);
 
-            // Core always adds the error queue as a "sending address" so it's likely already added hence the TryAdd 
+            // Core always adds the error queue as a "sending address" so it's likely already added hence the TryAdd
             queuesToCreate.TryAdd(errorQueueName, BuildDefaultCreateQueueOptions(errorQueueName));
 
             var receiveQueueName = AzureServiceBusTransportInfrastructure.ToTransportAddress(receiver.ReceiveAddress, DestinationManager);
@@ -293,7 +293,7 @@ public partial class AzureServiceBusTransport : TransportDefinition
     /// <summary>
     /// Gets or sets the maximum time period that a queue can remain idle before Azure Service Bus automatically deletes it.
     /// </summary>
-    /// <value>The idle timeout after which unused entities are automatically deleted. The minimum allowed value is 5 minutes.</value>    
+    /// <value>The idle timeout after which unused entities are automatically deleted. The minimum allowed value is 5 minutes.</value>
     /// <remarks>
     /// <para>
     /// This property controls the AutoDeleteOnIdle setting for queues created by the transport.
@@ -303,7 +303,7 @@ public partial class AzureServiceBusTransport : TransportDefinition
     /// </para>
     /// <para>
     /// This setting only applies to queues, not to topics or subscriptions. Topics and subscriptions are considered
-    /// shared infrastructure and are not affected by this property. Only instance-specific input queues (such when using 'MakeInstanceUniquelyAddressable') 
+    /// shared infrastructure and are not affected by this property. Only instance-specific input queues (such when using 'MakeInstanceUniquelyAddressable')
     /// will have AutoDeleteOnIdle applied, while shared queues (such as error and audit queues) remain unaffected to prevent unintended deletion of critical infrastructure.
     /// </para>
     /// <para>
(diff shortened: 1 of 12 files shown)


danielmarbach:
@SeanFeldman FYI

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -435,6 +435,30 @@ public IWebProxy? WebProxy
...
+    /// Set the maximum delivery count that is applied to queues that are created when the infrastructure is setup
+    /// </summary>
+    /// <remarks>This MaxDeliveryCount is set to <value>int.Max</value> by default, to stay backward compatible will change in the
+    /// next major version to a lower default value. When the emulator is used the value defaults to <value>10</value>
+    /// unless explicitly set to another value.
+    /// <para>
+    /// NServiceBus recoverability controls retry decisions via the recoverability policy, so this value should be high enough
+    /// to allow the recoverability policy to eventually move the message to the error queue. It should also not be set so high
+    /// that it effectively becomes infinite retries. To use the native dead letter queue instead of the error queue,
+    /// enable <see cref="AutoForwardDeadLetteredMessagesToErrorQueue"/>.
```
andreasohlund: AutoForwardDeadLetteredMessagesToErrorQueue enables the forwarding of the DLQ to the central error, so I think this needs to be reworded. Did you mean `MoveErrorsToAzureServiceBusDeadLetterQueue`?

danielmarbach: No the intent was to indicate when stuff lands in the DLQ because the value was not high enough. You probably want to set `AutoForwardDeadLetteredMessagesToErrorQueue`

andreasohlund: Ah, should we try to reword it along the lines of "When changing this to a lower value, we recommend to enable AutoForwardDeadLetteredMessagesToErrorQueue so that failures would become visible in the Platform"?

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -435,6 +435,30 @@ public IWebProxy? WebProxy
         }
     }
 
+    /// <summary>
+    /// Set the maximum delivery count that is applied to queues that are created when the infrastructure is setup
```
DavidBoike:
```suggestion
    /// Set the maximum delivery count that is applied to queues that are created when the infrastructure is setup.
```

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -435,6 +435,30 @@ public IWebProxy? WebProxy
...
+    /// </summary>
+    /// <remarks>This MaxDeliveryCount is set to <value>int.Max</value> by default, to stay backward compatible will change in the
+    /// next major version to a lower default value. When the emulator is used the value defaults to <value>10</value>
+    /// unless explicitly set to another value.
```
DavidBoike:
```suggestion
    /// <remarks>
    /// <para>
    /// This MaxDeliveryCount is set to <value>int.MaxValue</value> by default, to stay backward compatible. the
    /// This value will change to a lower default value in the next major version. 
    /// When the emulator is used the value defaults to <value>10</value>
    /// unless explicitly set to another value.
    /// </para>
```

src/Transport/AzureServiceBusTransport.cs
DavidBoike: Yes, and also maybe an explanation that _if/when_ you set the limit lower than recoverability math allows, it's going to go to the DLQ.

andreasohlund (approved):
Docs for this can be added to https://github.com/Particular/docs.particular.net/pull/8191

SeanFeldman:
Would it be possible to do the recoverability math and compare the result to the `MaxDeliveryCount` configured by the user, since both are transport settings, and if it's not high enough, log a warning, mentioning `AutoForwardDeadLetteredMessagesToErrorQueue`? While XML doco is great, having that in the user face as a log can help even more.

danielmarbach:
Hmm... Good suggestion

> Would it be possible to do the recoverability math and compare the result to the `MaxDeliveryCount` configured by the user, since both are transport settings, and if it's not high enough, log a warning, mentioning `AutoForwardDeadLetteredMessagesToErrorQueue`? While XML doco is great, having that in the user face as a log can help even more.

The problematic part is that the setting on the transport and what the queue actually has configured might not be in sync in all cases. I also wonder how we could determine "not high enough" because, effectively, the count is not that relevant. What is relevant is the decision outcome of the recovery action. We are currently logging when dead-lettering.

But maybe checking the MaxDeliveryCount local setting against DeliveryCount on the received message in case we decide to dead-letter and then log information based on that is enough? Although you could argue that for people that explicitly don't want to use `AutoForwardDeadLetteredMessagesToErrorQueue` this log statement could be annoying... I'm torn

SeanFeldman:
> I'm torn

Then sleep on it and see what makes sense. I'm just thinking that folks don't read every setting doco but they sure pay attention to warnings in production 🙂

danielmarbach:
@SeanFeldman I think what could work is that we could make the setting nullable. If it is not explicitly set, we log a warning and recommend it. If someone sets it to `false`, we stop logging. 

@andreasohlund thoughts?

andreasohlund:
Sounds resonable

danielmarbach:
@SeanFeldman done.

andreasohlund (approved):

src/Transport/AzureServiceBusTransport.cs
```diff
@@ -332,13 +332,27 @@ public TimeSpan? AutoDeleteOnIdle
...
+    /// When set to <see langword="false"/>, dead-lettered messages remain in the Azure Service Bus dead-letter queue and no warning is logged.
+    /// Use this value to explicitly opt out of both forwarding and the associated warning.
+    /// </para>
+    /// <para>
     /// This option only affects queues created by the transport during infrastructure setup. It applies to transport-created endpoint queues,
     /// including instance-specific queues, and excludes the error queue itself to avoid self-forwarding loops.
+    /// </para>
     /// </remarks>
-    public bool AutoForwardDeadLetteredMessagesToErrorQueue { get; set; }
+    public bool? AutoForwardDeadLetteredMessagesToErrorQueue { get; set; }
```
DavidBoike, line 355: Can we get away with this type change in a minor version?

andreasohlund: AutoForwardDeadLetteredMessagesToErrorQueue was added in v6.3 so not released yet

DavidBoike (approved):

PR #1401, opened 2026-04-22
Adjust how DLQ properties are upconverted to fault headers

andreasohlund:
This exposes the DLQ Reason as the "ExceptionType" and DLQ Description as "Message" to better align with assumptions in ServiceControl. This also exposes the MessageId, Header and Body extraction to allow the function package to reused the same logic.

### DIFF
--- a/src/Transport/Receiving/MessageExtensions.cs
+++ b/src/Transport/Receiving/MessageExtensions.cs
@@ -16,6 +16,10 @@ public static class MessageExtensions
     /// <summary>
     /// Extracts NServiceBus headers from the <see cref="ServiceBusReceivedMessage"/>
     /// </summary>
+    /// <remarks>
+    /// In addition to converting all application properties to headers native message properties like ReplyTo, ContentType are also exposed as properties.
+    /// For dead lettered messages source, reason and description is mapped to relevant NServiceBus fault headers for better integration with the platform.
+    /// </remarks>
     public static Dictionary<string, string?> GetNServiceBusHeaders(this ServiceBusReceivedMessage message)
     {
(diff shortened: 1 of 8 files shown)


src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -69,6 +78,9 @@ public static string GetMessageId(this ServiceBusReceivedMessage message)
         return GuidHelper.CreateVersion8(message.EnqueuedTime, message.SequenceNumber).ToString();
     }
 
+    /// <summary>
+    /// Returns the message body of the <see cref="ServiceBusReceivedMessage"/>.
+    /// </summary>
```
danielmarbach: I think the description should make clear what this does compared to just accessing the bodz property of the message

src/Transport/Receiving/MessageExtensions.cs
```diff
@@ -43,17 +49,20 @@ static class MessageExtensions
...
-            headers[FaultsHeaderKeys.StackTrace] = message.DeadLetterErrorDescription;
+            headers[FaultsHeaderKeys.Message] = message.DeadLetterErrorDescription;
         }
 
         return headers;
     }
 
+    /// <summary>
+    /// Returns the message ID for the <see cref="ServiceBusReceivedMessage"/>.
+    /// </summary>
```
danielmarbach: Same here we need to explain what this does compared to accessing the id now that it is public

andreasohlund: Good catch, all docs are now more detailed

danielmarbach (approved):

PR #1407, opened 2026-04-28
Fix flaky concurrency test by enforcing limit with own SemaphoreSlim

### LINKED ISSUE #1212: Transport Test "When_multiple_messages_are_available_and_concurrency_is_lowered_after_pump_started" is flaky on the build server
### Describe the bug

The following test behave flaky

* `When_multiple_messages_are_available_and_concurrency_is_lowered_after_pump_started`

### Steps to reproduce

Run the test on the CI build
(Note: This test is not currently being run for ASB because it is considered flaky)

### Relevant log output

_No response_

### Additional Information

It might be related to the Service Bus Processor trying to reconciliate the concurrency together with the prefetch count and that process can take some time.

We have switched to the native way of doing the concurrency and prefetch updates because the SDK already provides such a functionality (and is already in charge of concurrent execution and prefetching). The previous way of starting and stopping the infrastructure had additional overload of having to stop and recreate processors, which also meant more overhead because the underlying connection means had to be re-established by the SDK.

### Steps to reproduce

Run on the build server

### Relevant log output

```shell

```

mauroservienti:
## Problem

`When_multiple_messages_are_available_and_concurrency_is_lowered_after_pump_started` was flaky on CI (#1212). The test lowers the concurrency limit from 10 to 2 after the pump starts, sends messages, and asserts that at most 2 are processed concurrently.

The failure:
```
Assert.That(maximumConcurrentMessages, Is.EqualTo(newConcurrencyLevel))
  Expected: 2
  But was:  4
```

The root cause is in the Azure SDK's `ServiceBusProcessor.UpdateConcurrency`. Internally it uses fire-and-forget `WaitAsync()` calls to drain semaphore slots when reducing concurrency. This creates a race window — however brief — where incoming messages can be dispatched at the old concurrency level before the reduction takes effect.

## Solution

Added a `SemaphoreSlim` in `MessagePump` as a strict enforcement layer on top of the SDK's best-effort concurrency control.

- `ChangeConcurrency` now **awaits** slot draining before returning. When reducing concurrency with no in-flight handlers (the test scenario), the slots drain synchronously and the new limit is in effect the instant `ChangeConcurrency` returns. When handlers are in-flight, it waits for enough to complete before returning.
- `OnProcessMessage` acquires the semaphore before processing and releases it in a `finally` block — making it the hard gate on concurrency regardless of SDK timing.
- `UpdateProcessingCapacity` (the circuit breaker path) adjusts the semaphore fire-and-forget, which is appropriate since that path is a soft/best-effort limit and is called from within a lock.
- The SDK's `UpdateConcurrency`/`UpdatePrefetchCount` calls are kept as an alignment hint.

Closes #1212

### DIFF
--- a/src/Transport/Receiving/MessagePump.cs
+++ b/src/Transport/Receiving/MessagePump.cs
@@ -31,6 +31,8 @@ sealed class MessagePump(
     // Start
     CancellationTokenSource? messageProcessingCancellationTokenSource;
     ServiceBusProcessor? processor;
+    SemaphoreSlim? concurrencySemaphore;
+    int activeConcurrencyLimit;
 
     readonly ILog Logger = LogManager.GetLogger<MessagePump>();
 
@@ -83,6 +85,9 @@ sealed class MessagePump(
 
         messageProcessingCancellationTokenSource = new CancellationTokenSource();
 
+        activeConcurrencyLimit = limitations!.MaxConcurrency;
+        concurrencySemaphore = new SemaphoreSlim(activeConcurrencyLimit);
+
         circuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker($"'{receiveSettings.ReceiveAddress}'",
             transportSettings.TimeToWaitBeforeTriggeringCircuitBreaker, ex =>
             {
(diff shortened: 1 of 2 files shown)


danielmarbach:
I'm not convinced adding this complexity to get the test passing is worth it nor that we put so much thought into that Core test to declare it as "strictly enforcing those boundaries". I'm 👎 for merging this

When we did https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/639 the idea was to switch away from custom processing to rely on the higher-level primitives like ServiceBusProcessor. With that change we also switched towards relying on the built-in primitives to update the concurrency with all the consequences of that. With this PR we seem to be going halfway backwards toward taking back some manual control without more reasoning than making a test pass.

I don't think we lose much by keeping the test commented out. Alternatively, we could make the test assertions partial so that the transport can override those to be less strict or see if we can contribute this mode (if really needed) to the SDK

mauroservienti:
I have no preferences.

afprtclr (approved):
> With this PR we seem to be going halfway backwards toward taking back some manual control without more reasoning than making a test pass.

I tend to disagree with that. I think the current naming doesn't convey the vibe of "best-effort approach" enough and I would be surprised to see that I set the concurrency and it's not enforced (that breaks principle of least astonishment). I think it's more than "making a test pass". Especially that this may be used in an edgy path when the APIs can be throttled.

Regarding going halfway backwards - agreed, and it is still different that it used to be. We're now relying on the SDK to not cross the concurrency limit and we add another guardrail, so it's a cooperative approach (unless I misunderstood the code).

I reviewed this and it looks good to me.

tmasternak:
_Please treat this as potentially useful information from someone having context of the work on the endpoint throttling behavior_

The `ChangeConcurrency` mechanism was introduced on the `IMessageReciever` as part of [the work to support throttling when consecutive failures happen on an endpoint level](https://github.com/Particular/NServiceBus/commit/c7144ef27ec0d8a649d8b31a3cbbc51fe091a747#diff-23e282ba38f8fa6379a1600dc082ea65f5b1aac239c50cc5cd88ce371f61ac53R21-R25). The test was introduced later, but I assume this was a follow-up work.

From the user's perspective, the expected behavior of the throttling mechanism is this:
> When multiple consecutive failures happen, the endpoint will switch into low-concurrency processing mode (aka probing mode) that prevents us from ending up with lots of error messages, and enables detecting when things are back to normal by processing one message at a time. When things seem to be back to normal, the mechanism will ramp up the concurrency level to the original value.  

In light of these requirements, I would argue that "after calling `ChangeConcurrency`, the endpoint will **eventually** throttle down (or ramp up) to the given, target concurrency level" is good enough to handle this end-user requirement.

Secondly, I would argue that visible user behavior (things users can observe in their handler code) when the concurrency limit is changed, not in synchronous (changes applied after `await` call finishes) but asynchronously (eventually), is pretty minimal, if detectable at all.

In summary, I would argue that if anything might need changing, this would be the tests, which are currently testing stronger requirements than are actually needed by the throttling mechanism.

danielmarbach:
> convey the vibe of "best-effort approach" enough 

`When_multiple_messages_are_available_and_concurrency_is_lowered_after_pump_started` with `Fewer_messages_handled_concurrently` gives me that vibe, but I realize that is my interpretation.

Regardless of that, @tmasternak's comment summarizes very well my stance

danielmarbach:
Some more context. Should we decide synchronous behavior was the original intent, then stopping the receiver and starting the receiver (the implementation we moved away from in https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/639/changes#diff-bf785c1b37e39481e89856ea4195ca5c7e21f862b8eb2ffc43b63208a3c582e5L118-L123 towards the built-in behavior) would achieve that.

It could very well be that our interpretation of the expected behavior was wrong at the time, and we are looking at a bug. But I have a hunch the tests are green because our default choice was for quite a few transports to rely on stop and start because they had no built-in way to adjust concurrency.

I did a brief check and it looks like this

For transports using Stop/Start (Learning, RabbitMQ, SQS, ASQ): StopReceive drains all in-flight messages before returning, then StartReceive creates a fresh pump with the new concurrency --> all before `ChangeConcurrency` returns. Once the await completes, the new concurrency IS in effect atomically from the caller's perspective. The test assertion `maximumConcurrentMessages == newConcurrencyLevel` is valid for these four transports and matches the implementation details.

For transports NOT using Stop/Start (MSMQ, SQL Server, ASB): The race window is exists:
- MSMQ/SQL: The semaphore swap is atomic but in-flight messages on the old semaphore continue concurrently. maximumConcurrentMessages can briefly reach oldConcurrency + newConcurrency.
- ASB: Returns Task.CompletedTask immediately while the SDK drain is fire-and-forget async. Peak observed concurrency can be the old level.

We are probably not experiencing that race with MSQM and SQL because the swap + drain is "fast enough"

If my thinking is correct we effectively have

3 out of 7 transport behaving eventual consistent and 4 behaving consistent

afprtclr:
> When_multiple_messages_are_available_and_concurrency_is_lowered_after_pump_started with Fewer_messages_handled_concurrently gives me that vibe, but I realize that is my interpretation.

These are tests. I was referring to the vibe of the `ChangeConcurrency` method name. If it was `TryChangeConcurrency` or `BeginChangeConcurrency` or `ChangeConcurrencyEventually`, then that would give a hint that this is not an immediate operation.

> It could very well be that our interpretation of the expected behavior was wrong at the time, and we are looking at a bug

I wouldn't call it a bug, but it definitely breaks principle of least astonishment. It's like I call `File.Delete` and the file is still there but disappears only some time after. Not to mention, that I have no way of telling when the concurrency if changed _for real_ which I may need (e.g., for throttling or rate limiting).

mauroservienti:
It seems to me the preferable course of action could be:

- Close this PR without merging
- Close the corresponding issue linking to this PR alongside an explanation of why it was not tackled
- Keep the link in the csproj exclusion to help connect the dots
- Raise a documentation PR explaining that `ChangeConcurrency` in ASB is eventually consistent since that's the SDK's underlying behavior

Thoughts?

danielmarbach:
I did a quick check and the behavior of adjusting the concurrency asynchronously has been the standard since v3 of the transport.

https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/3.0.0/src/Transport/Receiving/MessagePump.cs#L221

Given we have no evidence this causes any problems and we have evidence of this discussion that the synchronous settlement of concurrency is not something that is really required (since other transports behave similarly) I'm leaning towards Closing this PR without merging, keep an explanation in the csproj why this test is excluded.

That being said, as far as I remember, @SzymonPobiega and @WilliamBZA have been part of the implementation efforts so they might be able to shed some light here and clarify whether the 3 transports that do the settlement async are correct or "incorrect" as by the previously intended design.

SzymonPobiega:
@danielmarbach the goal was to make sure that _eventually_ the concurrency is adjusted. When we implemented the feature, we were thinking about two alternatives that we ruled out in favor of the change concurrency API:
- Build the auto-rate-limiting feature outside of the endpoint i.e. stop the endpoint and start with different settings. This was ruled out because it would be quite difficult to use and potentially interfere with hosting abstractions
- Refactor the endpoint code and allow to replace the transport (with one with different concurrency settings) while an endpoint is in the running state. This was ruled out because it would generate significant complexity in the core

Instead, we decided to add a bit more complexity to each transport. In the hindsight, we probably should have thought more about immediate vs eventual nature of the concurrency level change and created a better test suite for it.

mauroservienti:
Withdrawn. More details in https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/1212#issuecomment-4398686054

PR #1421, opened 2026-05-13
Fix condition checks for DeadLetterReason and ErrorDescription

andreasohlund:
Since native DLQ messages won't have any exception headers at all, the code worked anyway but this aligns the check with the actual upconversion


danielmarbach (approved):

danielmarbach:
@andreasohlund worth patching this?

andreasohlund:
I think no since either all of none of these headers are present

PR #1456, opened 2026-07-02
[WIP] Support for ordered delivery (through ASB sessions) for Pub/Sub 

poornimanayar:


src/Transport/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -105,6 +105,12 @@ IMessageReceiver CreateMessagePump(ReceiveSettings receiveSettings, ServiceBusCl
                 throw new Exception("Cannot use a session-enabled receiver to receive from a dead-letter queue");
             }
 
+            //TODO : What about raw mode?
```
SzymonPobiega: I think we could use the number of `ReceiveSettings` objects in the collection to detect the send-only setup more reliably (regardless if hosted in core):
- More then 0 `ReceiveSettings` -> OK to `EnableSessions`
- Zero `ReceiveSettings` -> throw

SzymonPobiega:
Replaced by https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1493

PR #1457, opened 2026-07-08
Trimming too long properties

afprtclr:
Fix for https://github.com/Particular/NServiceBus/issues/5764

Trimming too long properties that would lead to failure when sending a message.

### DIFF
--- a/src/Transport/Sending/OutgoingMessageExtensions.cs
+++ b/src/Transport/Sending/OutgoingMessageExtensions.cs
@@ -8,7 +8,8 @@ using Azure.Messaging.ServiceBus;
 static class OutgoingMessageExtensions
 {
     // The actual property size limit is 32767 but the total size of all properties must be at most 65534.
-    // We use buffer to avoid hitting the limit
+    // Based on https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-quotas
+    // We also use some buffer to avoid hitting the limit
     const int MaxPropertySize = 32767;
     const int Buffer = 2048;
     static readonly string[] HeadersToTrim = ["NServiceBus.ExceptionInfo.StackTrace", "NServiceBus.ExceptionInfo.Message"];
--- a/src/Transport/Sending/OutgoingMessageExtensions.cs
+++ b/src/Transport/Sending/OutgoingMessageExtensions.cs
@@ -91,27 +91,28 @@ static class OutgoingMessageExtensions
         }
     }
 
-static void TrimTooLongKnownHeadersInPlace(ServiceBusMessage message)
-{
-    var maxUtf8Bytes = MaxPropertySize - Buffer;
-
-    Encoder? encoder = null;
-    byte[]? trimBuffer = null;
-
-    foreach (var headerName in HeadersToTrim)
+    static void TrimTooLongKnownHeadersInPlace(ServiceBusMessage message)
     {
-        if (!message.ApplicationProperties.TryGetValue(headerName, out var headerValue) ||
-            headerValue is not string value ||
-            Encoding.UTF8.GetByteCount(value) <= maxUtf8Bytes)
+        var maxUtf8Bytes = MaxPropertySize - Buffer;
+
+        Encoder? encoder = null;
+        byte[]? trimBuffer = null;
+
+        foreach (var headerName in HeadersToTrim)
         {
-            continue;
+            if (!message.ApplicationProperties.TryGetValue(headerName, out var headerValue) ||
+                headerValue is not string value ||
+                Encoding.UTF8.GetByteCount(value) <= maxUtf8Bytes)
+            {
+                continue;
+            }
+
+            encoder ??= Encoding.UTF8.GetEncoder();
+            trimBuffer ??= new byte[maxUtf8Bytes];
+
+            encoder.Reset();
+            encoder.Convert(value.AsSpan(), trimBuffer.AsSpan(), flush: false, out _, out int bytesUsed, out _);
+            message.ApplicationProperties[headerName] = Encoding.UTF8.GetString(trimBuffer, 0, bytesUsed);
         }
-
-        encoder ??= Encoding.UTF8.GetEncoder();
-        trimBuffer ??= new byte[maxUtf8Bytes];
-
-        encoder.Reset();
-        encoder.Convert(value.AsSpan(), trimBuffer.AsSpan(), flush: false, out _, out int bytesUsed, out _);
-        message.ApplicationProperties[headerName] = Encoding.UTF8.GetString(trimBuffer, 0, bytesUsed);
     }
 }
--- a/src/Transport/Sending/OutgoingMessageExtensions.cs
+++ b/src/Transport/Sending/OutgoingMessageExtensions.cs
@@ -2,7 +2,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Text;
 using Azure.Messaging.ServiceBus;
 
@@ -94,34 +93,21 @@ static class OutgoingMessageExtensions
 
     static void TrimTooLongKnownHeadersInPlace(ServiceBusMessage message)
     {
-        foreach (var headerNamer in HeadersToTrim)
+        var encoder = Encoding.UTF8.GetEncoder();
+        var trimBuffer = new byte[MaxPropertySize - Buffer];
+
+        foreach (var headerName in HeadersToTrim)
         {
-            message.ApplicationProperties.TryGetValue(headerNamer, out var headerValue);
+            message.ApplicationProperties.TryGetValue(headerName, out var headerValue);
 
             if (headerValue is not string value || Encoding.UTF8.GetByteCount(value) <= MaxPropertySize)
             {
                 continue;
             }
 
-            var sb = new StringBuilder(value.Length);
-            int usedBytes = 0;
-
-            var enumerator = StringInfo.GetTextElementEnumerator(value);
-            while (enumerator.MoveNext())
-            {
-                string element = (string)enumerator.Current;
-                int elementBytes = Encoding.UTF8.GetByteCount(element);
-
-                if (usedBytes + elementBytes + Buffer > MaxPropertySize)
-                {
-                    break;
-                }
-
-                sb.Append(element);
-                usedBytes += elementBytes;
-            }
-
-            message.ApplicationProperties[headerNamer] = sb.ToString();
+            encoder.Reset();
+            encoder.Convert(value.AsSpan(), trimBuffer.AsSpan(), flush: false, out _, out int bytesUsed, out _);
+            message.ApplicationProperties[headerName] = Encoding.UTF8.GetString(trimBuffer, 0, bytesUsed);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -81,4 +91,37 @@ static void CopyHeaders(ServiceBusMessage outgoingMessage, Dictionary<string, st
             outgoingMessage.ApplicationProperties[header.Key] = header.Value;
         }
     }
+
+    static void TrimTooLongKnownHeadersInPlace(ServiceBusMessage message)
+    {
+        foreach (var headerNamer in HeadersToTrim)
```
mnadareski-particular: Nitpick: `headerName` instead of `headerNamer`

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -81,4 +91,37 @@ static void CopyHeaders(ServiceBusMessage outgoingMessage, Dictionary<string, st
             outgoingMessage.ApplicationProperties[header.Key] = header.Value;
         }
     }
+
+    static void TrimTooLongKnownHeadersInPlace(ServiceBusMessage message)
```
mnadareski-particular, line 95: In some research, I found two possible alternative approaches to the per-character building that is done here:

- Working from the total length of the existing string and using progressively shorter substrings for calcuation. The upside is that the code has fewer lines using both an explicit loop and a Linq query. The downside is that it may be less performant given the upper bound on string length is greater than the upper bound imposed on the header value.
- Using [Encoder.Convert](https://learn.microsoft.com/en-us/dotnet/api/system.text.encoder.convert?view=net-10.0), which allows for the ability to limit the output by both input number of characters and output number of bytes. The upside is that it only takes two lines (one for getting the encoder specific to the encoding we're using and one to call the encoder). The downside is that there are many output parameters we would be ignoring for most cases.

I think that trying out the `Encoder.Convert` variant would be a good choice since it has the possibility to be more performant and use a built-in method to reduce complexity.

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -81,4 +91,37 @@ static void CopyHeaders(ServiceBusMessage outgoingMessage, Dictionary<string, st
...
+                if (usedBytes + elementBytes + Buffer > MaxPropertySize)
+                {
+                    break;
+                }
+
+                sb.Append(element);
+                usedBytes += elementBytes;
+            }
+
+            message.ApplicationProperties[headerNamer] = sb.ToString();
```
mnadareski-particular: If using `Encoder.Convert`, this code would end up looking like this:
```suggestion
            var encoder = Encoding.UTF8.GetEncoder();
            
            byte[] arr = new byte[MaxPropertySize - Buffer];
            encoder.Convert(value.AsSpan(), arr.AsSpan(), flush: false, out _, out int bytes, out _);
            
            message.ApplicationProperties[headerNamer] = Encoding.UTF8.GetString(arr, 0, bytes);
```

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -2,10 +2,18 @@
...
+using System.Globalization;
+using System.Text;
 using Azure.Messaging.ServiceBus;
 
 static class OutgoingMessageExtensions
 {
+    // The actual property size limit is 32767 but the total size of all properties must be at most 65534.
+    // We use buffer to avoid hitting the limit
+    internal const int MaxPropertySize = 32767;
+    const int Buffer = 2048;
```
mnadareski-particular, line 14: If using `Encoder.Convert`, then the `Buffer` constant is no longer needed.

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -2,10 +2,18 @@
...
 using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
 using Azure.Messaging.ServiceBus;
 
 static class OutgoingMessageExtensions
 {
+    // The actual property size limit is 32767 but the total size of all properties must be at most 65534.
+    // We use buffer to avoid hitting the limit
+    internal const int MaxPropertySize = 32767;
```
mnadareski-particular: I mentioned this in our meeting, but I think it would be good for me to comment here for posterity as well:

In previous positions, the use of constants from the main logic was discouraged to use in the tests. The rationale is that if the constant changed for some reason, the tests would get whatever the new value is, regardless if that value was correct now or not.

I do not feel strongly about the use of constants in the tests and I will not request that this change if this would be the only change, but it may be something to consider in future testing.

src/Transport/Sending/OutgoingMessageExtensions.cs
afprtclr: It is. We don't want to trim to the MaxPropertySize but rather lave some buffer.

mnadareski-particular:
I've confirmed locally that the change to use the encoder seems to work the same way that the current code does. I had to tweak my suggested code to take into account the buffer size.

I also noted in my testing that the encoder could probably be set outside of the loop and just reused for each loop iteration as I don't believe anything that we're doing would change the internal state of it.

afprtclr:
@copilot Please apply changes described by @mnadareski-particular  in this PR.

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -81,4 +90,24 @@ static void CopyHeaders(ServiceBusMessage outgoingMessage, Dictionary<string, st
             outgoingMessage.ApplicationProperties[header.Key] = header.Value;
         }
     }
+
+    static void TrimTooLongKnownHeadersInPlace(ServiceBusMessage message)
+    {
+        var encoder = Encoding.UTF8.GetEncoder();
```
afprtclr: @copilot Analyze if this variable can be moved to a field of the class and initialize it once.

src/Transport/Sending/OutgoingMessageExtensions.cs
afprtclr: @copilot Please make MaxPropertySize and HeadersToTrim  private (remove the internal modifier), and hardcode these constants in the tests.

src/Transport/Sending/OutgoingMessageExtensions.cs
afprtclr: Cool.

src/Transport/Sending/OutgoingMessageExtensions.cs
afprtclr: Fixed.

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -29,6 +36,8 @@ public static ServiceBusMessage ToAzureServiceBusMessage(
 
         CopyHeaders(message, outgoingMessage.Headers);
 
+        TrimTooLongKnownHeadersInPlace(message);
+
```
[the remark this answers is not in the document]
afprtclr: This is good enough.

mnadareski-particular (approved):

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -2,10 +2,17 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text;
 using Azure.Messaging.ServiceBus;
 
 static class OutgoingMessageExtensions
 {
+    // The actual property size limit is 32767 but the total size of all properties must be at most 65534.
```
bording, line 10: What is the source of this information? Would be helpful to have a link in the comment if it's documented somewhere.

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -81,4 +90,29 @@ static void CopyHeaders(ServiceBusMessage outgoingMessage, Dictionary<string, st
...
+                Encoding.UTF8.GetByteCount(value) <= maxUtf8Bytes)
+            {
+                continue;
+            }
+
+            encoder ??= Encoding.UTF8.GetEncoder();
+            trimBuffer ??= new byte[maxUtf8Bytes];
+
+            encoder.Reset();
+            encoder.Convert(value.AsSpan(), trimBuffer.AsSpan(), flush: false, out _, out int bytesUsed, out _);
```
bording, line 115: If we set `flush` to `true`, could we get rid of the `Reset` call?

bording (commented):
Overall this does seem to address the specific concern of the issue, which was that a single message property was too large.

I'm wondering if we should be thinking more about the overall size limit as well, since it certainly seems like it would be possible to run up against that still here (Even with that `Buffer`)...

Or maybe we worry about that later.

src/Transport/Sending/OutgoingMessageExtensions.cs
afprtclr: @copilot Please verify if that is possible. If yes, then update the code.

src/Transport/Sending/OutgoingMessageExtensions.cs
afprtclr: Added a link to the docs.

afprtclr:
> I'm wondering if we should be thinking more about the overall size limit as well, since it certainly seems like it would be possible to run up against that still here

Yeah, we discussed that and decided to not do that. We wouldn't know which properties to trim anyway. ASQ has a fallback logic to send the message as-is on exception, maybe something like that should be implemented one day in ASB.

src/Transport/Sending/OutgoingMessageExtensions.cs
```diff
@@ -81,4 +91,29 @@ static void CopyHeaders(ServiceBusMessage outgoingMessage, Dictionary<string, st
...
+        var maxUtf8Bytes = MaxPropertySize - Buffer;
+
+        Encoder? encoder = null;
+        byte[]? trimBuffer = null;
+
+        foreach (var headerName in HeadersToTrim)
+        {
+            if (!message.ApplicationProperties.TryGetValue(headerName, out var headerValue) ||
+                headerValue is not string value ||
+                Encoding.UTF8.GetByteCount(value) <= maxUtf8Bytes)
```
danielmarbach, line 106: Given this now has to scan and count every header that matches would it be worthwhile to do a cheap prescan first?

Something like 

if (value.Length <= maxPropertySize / 3)
{
    continue;
}

And then you might not even need the exact byte count and could fall back to convert directly by checking the completed flag

And given exceptions may occur in bursts up to the concurrency of an endpoint would lazy using a pooled array be worthwhile too that is returned on finally (probably without clearing)?

afprtclr: > Given this now has to scan and count every header that matches would it be worthwhile to do a cheap prescan first?

Maybe, hard to say for the real scenarios. This happens for errors only, so should be quite rare in proper production scenarios.

> And given exceptions may occur in bursts up to the concurrency of an endpoint would lazy using a pooled array be worthwhile too that is returned on finally (probably without clearing)?

Maybe, we don't have any data points on that.

PR #1483, opened 2026-08-04
Batch isolated messages per destination and don't enlist in the existing transaction

jpalac:
Addresses https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/857

### DIFF
--- a/src/AcceptanceTests/Sending/When_batching_multiple_outgoing_small_messages.cs
+++ b/src/AcceptanceTests/Sending/When_batching_multiple_outgoing_small_messages.cs
@@ -75,7 +75,7 @@ public class When_batching_multiple_outgoing_small_messages : NServiceBusAccepta
         var singleBatchLines = logoutput.ReplaceLineEndings()
             .Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries)
             .Where(x => x.StartsWith("Sent batch '1' with '1'") && !x.Contains(kickOffMessageId));
-        Assert.That(singleBatchLines.Count(), Is.EqualTo(context.MessageIdsForImmediateDispatch.Count), "Should have used 1 batch for each immediate dispatch messages but didn't");
+        Assert.That(singleBatchLines.Count(), Is.EqualTo(context.MessageIdsForImmediateDispatch.Count), "Should have used 1 batch for each immediate dispatch message but didn't");
     }
 
     static string AggregateBatchLogOutput(ScenarioContext context)
--- a/src/AcceptanceTests/Sending/When_batching_multiple_outgoing_small_messages.cs
+++ b/src/AcceptanceTests/Sending/When_batching_multiple_outgoing_small_messages.cs
@@ -69,7 +69,7 @@ public class When_batching_multiple_outgoing_small_messages : NServiceBusAccepta
         }
 
         //Assert immediate dispatch messages were not batched together
-        Assert.That(logoutput, Does.Not.Contain($"Sent batch '1' with '{context.MessageIdsForImmediateDispatch.Count}'"), "Should have used 1 batch for each immediate dispatch messages but didn't");
+        Assert.That(logoutput, Does.Not.Contain($"Sent batch '1' with '{context.MessageIdsForImmediateDispatch.Count}'"), "Should have used 1 batch for each immediate dispatch message but didn't");
 
         //assert that immediate dispatch messages were sent in their own batches
         var singleBatchLines = logoutput.ReplaceLineEndings()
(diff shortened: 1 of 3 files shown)


src/AcceptanceTests/Sending/When_batching_multiple_outgoing_small_messages.cs
```diff
@@ -54,18 +54,28 @@ public async Task Should_batch_as_many_as_possible_for_non_immediate_dispatches(
...
         }
+
+        //Assert immediate dispatch messages were not batched together
+        Assert.That(logoutput, Does.Not.Contain($"Sent batch '1' with '{context.MessageIdsForImmediateDispatch.Count}'"), "Should have used 1 batch for each immediate dispatch messages but didn't");
+
+        //assert that immediate dispatch messages were sent in their own batches
+        var singleBatchLines = logoutput.ReplaceLineEndings()
+            .Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries)
+            .Where(x => x.StartsWith("Sent batch '1' with '1'") && !x.Contains(kickOffMessageId));
+        Assert.That(singleBatchLines.Count(), Is.EqualTo(context.MessageIdsForImmediateDispatch.Count), "Should have used 1 batch for each immediate dispatch messages but didn't");
```
stevedowling:
```suggestion
        Assert.That(singleBatchLines.Count(), Is.EqualTo(context.MessageIdsForImmediateDispatch.Count), "Should have used 1 batch for each immediate dispatch message but didn't");
```

src/AcceptanceTests/Sending/When_batching_multiple_outgoing_small_messages.cs
```diff
@@ -54,18 +54,28 @@ public async Task Should_batch_as_many_as_possible_for_non_immediate_dispatches(
...
         }
 
         foreach (var messageIdForImmediateDispatch in listOfMessagesForImmediateDispatch)
         {
-            Assert.That(logoutput, Does.Not.Contain(messageIdForImmediateDispatch), $"{messageIdForImmediateDispatch} should not be included in any of the batches. Output: {logoutput}");
+            Assert.That(logoutput, Does.Contain(messageIdForImmediateDispatch), $"{messageIdForImmediateDispatch} not found in any of the batches. Output: {logoutput}");
         }
+
+        //Assert immediate dispatch messages were not batched together
+        Assert.That(logoutput, Does.Not.Contain($"Sent batch '1' with '{context.MessageIdsForImmediateDispatch.Count}'"), "Should have used 1 batch for each immediate dispatch messages but didn't");
```
stevedowling:
```suggestion
        Assert.That(logoutput, Does.Not.Contain($"Sent batch '1' with '{context.MessageIdsForImmediateDispatch.Count}'"), "Should have used 1 batch for each immediate dispatch message but didn't");
```

src/Tests/Sending/MessageDispatcherTests.cs
```diff
@@ -196,11 +200,11 @@ public void Should_throw_when_multicast_dispatch_destination_not_available_and_t
                 [],
                 DispatchConsistency.Default);
 
-        Assert.That(async () => await dispatcher.Dispatch(new TransportOperations(operation1, operation2), new TransportTransaction()), Throws.Exception.InstanceOf<ServiceBusException>());
+        Assert.That(async () => await dispatcher.Dispatch(new TransportOperations(operation1, operation2), new TransportTransaction()), Throws.InstanceOf<InvalidOperationException>());
```
stevedowling: Why did it throw a `ServiceBusException` before and a `InvalidOperationException` now?

stevedowling (approved):

src/Tests/Sending/MessageDispatcherTests.cs
jasontaylordev: This could break existing code. Anyone explicitly catching `ServiceBusException` will stop catching it.

jpalac: Hmm - the path has changed for the isolated messages to go via the DispatchBatchOrFallbackToIndividualSendsForDestination method and there are inconsistencies in our code.

The code has this for the same error reason:
- DispatchForDestination (old path, [line 365](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/Sending/MessageDispatcher.cs#L360-L363)): throw; — re-throws the original ServiceBusException
- DispatchBatchOrFallbackToIndividualSendsForDestination (new path,[ line 256](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/Sending/MessageDispatcher.cs#L254-L257)): throw new InvalidOperationException(...)  -  wraps the ServiceBusException in an InvalidOperationException

The batch path was already wrapping the exception for default  consistency dispatches. Now that isolated dispatches also flow through the batch path, they get the wrapped exception too. 

They should be consistent as it's throwing an error for the same reason. I agree that it can break people - not sure how to get out of this though.

stevedowling (approved):

jasontaylordev (approved):

danielmarbach (approved):
More of a thought, and quite likely we don't need to address this:

Would we see any benefit from providing an opting-out option? Requires immediate dispatch wouldn't work because that changes the semantics when used with the outbox.

PR #1493, opened 2026-08-25
Add Publish/Subscribe to session-enabled mode using endpoint-driven forwarding approach

SzymonPobiega:
Replaces https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1456

Adds support for publish-subscribe pattern to session-enabled endpoints. Non-session-enabled endpoints rely on Azure Service Bus automatic forwarding from a subscription to the endpoint's input queue to implement publish-subscribe.

This approach is not possible for session-enabled endpoints because automatic forwarding cannot be enabled between session-enabled entities. Microsoft does not disclose the reason for it but it might be related to difficulty in preserving the order of messages being forwarded without hurting the scalability of the infrastructure.

This PR introduces endpoint-driven forwarding as a solution to the challenge. Each endpoint runs a number of forwarders (one per subscription) that use `ServiceBusProcessor` to move messages from a subscription to the input queue so that the behavior is preserved from the user's perspective.

The forwarders are fitted with circuit breakers to make sure persistent failures are correctly escalated (e.g. when the source subscription is deleted or the forwarding is misconfigured). There are not know situations in which an individual message might not be forwardable (i.e. _poison_) so no _poison_ message detection and handling logic is implemented.

The forwarding is configured via `Subscribe` and `Unsubscribe` APIs of the `IEndpointInstance` with following consequences:
- by default session-enabled subscribers require no additional configuration compared to non-session-enabled ones (auto-subscribe feature is active)
- when infrastructure is scripted and provisioned outside of NServiceBus, auto-subscribe has to be enabled for the subscribing endpoint to be able to process events. Alternatively, users might want to disable auto-subscribe and configure the subscriptions manually via `Subscribe` APIs of `IEndpointInstance`


src/Transport/Receiving/OrderedSubscriptionForwarder.cs
```diff
@@ -0,0 +1,156 @@
...
+using Logging;
+
+class OrderedSubscriptionForwarder : IAsyncDisposable
+{
+    static readonly ILog Logger = LogManager.GetLogger<OrderedSubscriptionForwarder>();
+
+    HashSet<string> eventTypes = [];
+    ServiceBusSessionProcessor? sessionProcessor;
+    ServiceBusSender? sender;
+    CancellationTokenSource forwardingCancellationTokenSource = new();
```
lailabougria: This is used in send and complete, but what about stop or dispose?

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            serviceBusMessage.ApplicationProperties.Add(messageApplicationProperty.Key, messageApplicationProperty.Value);
+        }
+
+        await arg.CompleteMessageAsync(arg.Message, forwardingCancellationTokenSource.Token).ConfigureAwait(false);
+        sender = forwardingClient.CreateSender(inputQueueAddress, new ServiceBusSenderOptions
+        {
+            Identifier = $"Forwarding-Sender-{topicName}-{subscriptionName}"
+        });
+        await sender.SendMessageAsync(serviceBusMessage, forwardingCancellationTokenSource.Token).ConfigureAwait(false);
```
lailabougria: It's wrapped in a transaction, but still i'd rather create the sender first, send the copy of the message before completing the incoming one.

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
```diff
@@ -0,0 +1,156 @@
...
+            To = arg.Message.To
+        };
+
+        foreach (var messageApplicationProperty in arg.Message.ApplicationProperties)
+        {
+            serviceBusMessage.ApplicationProperties.Add(messageApplicationProperty.Key, messageApplicationProperty.Value);
+        }
+
+        await arg.CompleteMessageAsync(arg.Message, forwardingCancellationTokenSource.Token).ConfigureAwait(false);
+        sender = forwardingClient.CreateSender(inputQueueAddress, new ServiceBusSenderOptions
```
lailabougria: We now create a sender per message we receive, should we be [caching](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebussender?view=azure-dotnet#remarks) them instead?

lailabougria: Also, they're not disposed anywhere atm

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
```diff
@@ -0,0 +1,156 @@
...
+                criticalErrorAction("Failed to receive message from Azure Service Bus.", ex,
+                    CancellationToken.None);
+            });
+    }
+
+    public async Task Start(CancellationToken cancellationToken = default)
+    {
+        var sessionReceiveOptions = new ServiceBusSessionProcessorOptions
+        {
+            PrefetchCount = 50, // TODO: do we want to make the prefetch count configurable
```
lailabougria: A scan for todo's is necessary, there's a couple

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
SzymonPobiega: 👍 we can initialize it when starting up

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
SzymonPobiega: I am pretty sure that for the send-via to work we first need to invoke the receiving queue in [this sample](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-transactions#see-it-in-code). Only then the transfer queue is associated with the client.

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
SzymonPobiega: Fixed. The processor's args token should be used instead of this one.

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
SzymonPobiega: Fixed. I put in safe defaults. No evidence we need configurable values (so far)

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
SzymonPobiega: Fixed

src/Transport/PubishOptionsExtensions.cs
```diff
@@ -0,0 +1,24 @@
+namespace NServiceBus;
+
+using System;
+using Extensibility;
+
+/// <summary>
+/// Extensions for <see cref="PublishOptions"/>.
+/// </summary>
+public static class PubishOptionsExtensions
```
lailabougria, line 9: Typo

src/Transport/PubishOptionsExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    /// <param name="options">PublishOptions</param>
+    /// <param name="sessionId">The session id to add to the dispatch properties</param>
+    /// <exception cref="ArgumentException"></exception>
+    public static void SetSessionId(this PublishOptions options, string sessionId)
+    {
+        ArgumentException.ThrowIfNullOrEmpty(sessionId);
+
+        var dispatchProperties = options.GetDispatchProperties();
+        dispatchProperties.TryAdd("SessionId", sessionId);
```
lailabougria, line 22: Do we want to at least warn if it was already set to another value?

src/Transport/EventRouting/MigrationTopologySubscriptionManager.cs
```diff
@@ -177,4 +177,6 @@ await CreationOptions.AdministrationClient.CreateSubscriptionAsync(subscription,
             Logger.Info($"Default subscription rule for topic {subscription.TopicName} is already in progress");
         }
     }
+
+    public override Task Shutdown(CancellationToken cancellationToken = default) => Task.CompletedTask;
```
lailabougria, line 181: Given that we don't support forwarding in this topology, should we then also throw on endpoint startup if enable sessions = true and topology = migration?

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
```diff
@@ -0,0 +1,163 @@
...
+            To = arg.Message.To
+        };
+
+        foreach (var messageApplicationProperty in arg.Message.ApplicationProperties)
+        {
+            serviceBusMessage.ApplicationProperties.Add(messageApplicationProperty.Key, messageApplicationProperty.Value);
+        }
+
+        await arg.CompleteMessageAsync(arg.Message, arg.CancellationToken).ConfigureAwait(false);
+        await sender!.SendMessageAsync(serviceBusMessage, arg.CancellationToken).ConfigureAwait(false);
```
lailabougria, line 111: We're doing this one by one, without considering batching as an option. Do we want to revisit that?

src/Transport/EventRouting/MigrationTopologySubscriptionManager.cs
SzymonPobiega: Yes, that or we can add support to it. I am leaning towards throwing an exception.

src/Transport/Receiving/OrderedSubscriptionForwarder.cs
SzymonPobiega: Not sure I understand. The SDK gives us one incoming message at the time.

src/Transport/PubishOptionsExtensions.cs
SzymonPobiega: I don't think that's what other similar APIs do (e.g. setting a header)

src/Transport/EventRouting/MigrationTopologySubscriptionManager.cs
SzymonPobiega: I pushed a fix for that

PR #1503, opened 2026-09-07
Introduce access to session state

lailabougria:


src/Transport/Receiving/SessionState/ISessionStateStore.cs
```diff
@@ -0,0 +1,22 @@
+namespace NServiceBus.Transport.AzureServiceBus;
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Messaging.ServiceBus;
+
+interface ISessionStateStore
```
SzymonPobiega, line 8: Do we need this internal interface? I can only spot a single implementation of it

src/Transport/Receiving/SessionState/AzureServiceBusSessionState.cs
```diff
@@ -0,0 +1,87 @@
...
+    public async Task Set<T>(T state, CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(state);
+
+        var current = await Load(cancellationToken).ConfigureAwait(false);
+
+        current.UserState = new UserSessionState
+        {
+            ContentType = "application/json",
+            Type = typeof(T).FullName,
```
SzymonPobiega, line 43: Do we need to store this? What's the expected usage?

src/Transport/Receiving/SessionState/AzureServiceBusSessionState.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    public async Task Set<T>(T state, CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(state);
+
+        var current = await Load(cancellationToken).ConfigureAwait(false);
+
+        current.UserState = new UserSessionState
+        {
+            ContentType = "application/json",
```
SzymonPobiega, line 42: If the `Data` is defined as `JsonElement`, can `ContentType` be anything other than json?

src/Transport/Receiving/SessionState/AzureServiceBusSessionState.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        current.UserState = null;
+        dirty = true;
+    }
+
+    /// <summary>
+    /// Persists pending session-state changes, if any. Called by the session pump right before the message is completed.
+    /// </summary>
+    public async Task Flush(CancellationToken cancellationToken = default)
```
SzymonPobiega, line 65: Do we want have the Save-Flush model enforced in our API?

It seems to be appropriate if the underlying `SetSessionStateAsync` operations are atomic with the message settlement but currently in our transport they are atomic only when using `SendsAtomicWithReceive` (default) transaction mode. When using `ReceiveOnly`, there is no `TransactionScope` surrounding the ASB API calls and as a result the `SetSessionStateAsync` is committed separately from `Complete`/`Abandon`.

As a result, we might get into some interesting gotachas:
- With outbox disabled, the session might be updated but the message fails to complete and comes back
- With outbox enabled, the message might be marked as processed in the outbox but later the `Flush` might fail and the session state update associated with the message handling is lost

Should we limit the session-state API to the `SendsAtomicWithReceive` mode (at least in the first release) to avoid these anomalies?

SzymonPobiega (changes requested):
I added some comments. Some are purely esthetic but the comment on the `Flash` method is more interesting -- documents transactional anomalies that we are exposed to.

src/Transport/Receiving/SessionState/AzureServiceBusSessionState.cs
lailabougria: I was thinking that this could be emited in observability

PR #1508, opened 2026-09-10
WIP adding activity label about sessions

poornimanayar:


src/Transport/Receiving/SessionsEnabledMessagePump.cs
```diff
@@ -261,6 +262,12 @@ async Task ProcessMessage(ServiceBusReceivedMessage message,
             using var azureServiceBusTransaction = CreateTransaction(message.PartitionKey);
             var messageContext = new MessageContext(nativeMessageId, headers, body, receiveProperties, azureServiceBusTransaction.TransportTransaction, ReceiveAddress, contextBag);
 
+            if (Activity.Current is { } activity)
+            {
+                messageContext.Extensions.Set(activity);
```
SzymonPobiega: Do we want to set the activity here in the context? How would that affect the core behavior?

poornimanayar: https://github.com/Particular/Platform/blob/main/design-decisions/open-telemetry/archive/traces-2022-07-15.md#decision is what I am referring to here and I have also seen the same pattern being used in IBM MQ transport. But on another look, this might not be needed at all, after all. Activity.Current might be enough as the MainPipelineExecutor starts an activity on the incoming pipeline

src/Transport/Receiving/SessionsEnabledMessagePump.cs
```diff
@@ -261,6 +262,11 @@ async Task ProcessMessage(ServiceBusReceivedMessage message,
             using var azureServiceBusTransaction = CreateTransaction(message.PartitionKey);
             var messageContext = new MessageContext(nativeMessageId, headers, body, receiveProperties, azureServiceBusTransaction.TransportTransaction, ReceiveAddress, contextBag);
 
+            if (Activity.Current is { } activity)
+            {
+                activity.AddTag("nservicebus.azureservicebus.session_id", message.SessionId);
```
lailabougria, line 267: If there's an activity here (that is only the case if the user opted in to the native SDK telemetry) then this will be fine. I'm thinking about the scenario in which that is not the case. In that case, with only this code, the session ID would be nowhere to be found, which feels off. Because of that, I would consider creating an activity in the transport if none exists (we'll need a dedicated ActivitySource, which will be automatically added for users who are adding NServiceBus.*) that captures the transport activity and adds this tag.

src/Transport/Receiving/SessionsEnabledMessagePump.cs
```diff
@@ -342,11 +348,7 @@ await processMessageEventArgs.SafeAbandonMessage(message,
     AzureServiceBusTransportTransaction CreateTransaction(string incomingQueuePartitionKey) =>
         TransactionMode == TransportTransactionMode.SendsAtomicWithReceive
             ? new AzureServiceBusTransportTransaction(receiveClient, incomingQueuePartitionKey,
-                new TransactionOptions
-                {
-                    IsolationLevel = IsolationLevel.Serializable,
-                    Timeout = TransactionManager.DefaultTimeout
-                })
+                new TransactionOptions { IsolationLevel = IsolationLevel.Serializable, Timeout = TransactionManager.DefaultTimeout })
```
lailabougria, line 351: Personal nit, but I found the expanded version easier to read

src/Transport/Receiving/SessionsEnabledMessagePump.cs
SzymonPobiega: Wouldn't that change the way the core works with regards to parent-child vs links? As far as I remember if the core sees an activity passed from the transport, it grabs it and bypasses its parent-child vs link logic entirely.

poornimanayar: Yes. I think that's what they have done in IBM MQ transport

